PR #18, opened 2024-03-04
Added github actions build for windows/linux/macos

gigi81:
Hope you will find this useful


gigi81:
You can see a sample build here https://github.com/gigi81/sourcegit/actions/runs/8133152657

ZCShou:
I like this PR. But for Linux, it is built using ubuntu-latest. I have some questions.
1. Do the ubuntu-latest mean Ubuntu 22.04？
2. Can it built using ubuntu-latest work for the other ubuntu?

ZCShou:
> I like this PR. But for Linux, it is built using ubuntu-latest. I have some questions.
> 
> 1. Do the ubuntu-latest mean Ubuntu 22.04？
> 2. Can it built using ubuntu-latest work for the other ubuntu?

I find [this](https://github.com/actions/runner-images), ubuntu-latest is Ubuntu 22.04

love-linger:
Very nice job. While the Linux port isn't complete at the moment, I'll merge this commit first.

gigi81:
The build that runs on Ubuntu produces a binary for linux-x64 which should run on any linux x64 in theory, so not only Ubuntu. I think that even if the app doesn't run or it is buggy on linux, having a build would help people to try it out and report bugs. There are ways to generate rpm or deb packages if you are interested I might be able to help on that but I think this is more flexible as it is like a portable build

PR #33, opened 2024-03-28
Open in...

ennerperez:
Contextual menu to "Open with actions..."

image

- Open in Fleet
- Open in Visual Studio Code


love-linger:
I've tested your PR. The functionality is complete. 

But using the `Avalonia.Svg` increase the size for final exe from 37MB to 42MB! Then I converts these two SVG icon to PNG image with 64x64 size and removed reference for `Avalonia.SVG`. After that the size of 
 final exe comes back to 37MB. 

So, I suggest use PNG instead of SVG for this PR.

ennerperez:
There is a file size limit for the final exe? I think the usage of Avalonia.Svg worth it. specially to improve the UI, add new icons, and make it scalable for different resolutions, I have to test it in high DPI resolution when probably PNG quality is not good enough.

love-linger:
> There is a file size limit for the final exe? I think the usage of Avalonia.Svg worth it. specially to improve the UI, add new icons, and make it scalable for different resolutions, I have to test it in high DPI resolution when probably PNG quality is not good enough.

No, there's no limitations for the size of final exe. But in this project, SVG only used to display the external editor icons. I think it's not worth to introduce a new package since we can use `RenderOptions.BitmapInterpolationMode="HighQuality"` to get a similar result with built-in PNG support.

ennerperez:
What happened with the menu?

love-linger:
Now, only external tools detected by the software are displayed.

PR #34, opened 2024-03-28
Enhance file and folder selection

walterlv:
Enhance file and folder selection by utilizing the default file manager instead of enforcing the use of explorer.exe.

Before this PR:

- file and folder selection was enforced to use explorer.exe.

After this PR:

- file and folder selection will use the default file manager.

Note:

I'm using Directory Opus as my default file manager, so I will be very happy if this PR is merged.


love-linger:
Thanks for the nice PR!

PR #37, opened 2024-03-28
Use system API to get default font name

walterlv:
Use system API to get default font name instead of setting a constant value.

These are the default font name for each culture:

| Language                          | Default font name  |
| --------------------------------- | ------------------ |
| 日语（Japanese）                  | Meiryo             |
| 韩语（Korean）                    | Malgun Gothic      |
| 繁体中文（Chinese (Traditional)） | Microsoft JhengHei |
| 简体中文（Chinese (Simplified)）  | Microsoft YaHei    |
| 希伯来语（Hebrew）                | Gisha              |
| 泰语（Thai）                      | Leelawadee         |

See my blog for more details:

- [Windows 系统的默认字体是什么？应用的默认字体是什么？](https://blog.walterlv.com/post/windows-default-font-family.html)

### DIFF
--- a/src/SourceGit/Native/Windows.cs
+++ b/src/SourceGit/Native/Windows.cs
@@ -60,7 +60,11 @@ namespace SourceGit.Native
             builder.With(new FontManagerOptions()
             {
                 DefaultFamilyName = SystemParameters.GetSystemDefaultFontName() ?? "Segoe UI",
-                FontFallbacks = [new FontFallback { FontFamily = "Segoe UI" }],
+                FontFallbacks =
+                [
+                    new FontFallback { FontFamily = "Segoe UI" },
+                    new FontFallback { FontFamily = "Tahoma" },
+                ],
             });
 
             // Fix drop shadow issue on Windows 10
--- a/src/SourceGit/Native/Windows.cs
+++ b/src/SourceGit/Native/Windows.cs
@@ -9,6 +9,8 @@ using Avalonia;
 using Avalonia.Controls;
 using Avalonia.Media;
 
+using SourceGit.Native.WindowsNative;
+
 namespace SourceGit.Native
 {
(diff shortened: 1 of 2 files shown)


walterlv:
It's very strange that the AvaloniaUI doesn't provide a way to get the default font name for each culture. Maybe I could try a PR for that.

love-linger:
It does NOT mean to get the system's default font, but the default font prefered by this app. For example: 

* Has this PR #37
  image
* Without this PR
  image

It's better when using JetBrains Mono with Microsoft YaHei UI as fallback font.

walterlv:
I've already escalated this issue to Microsoft for further attention. You can reference the reported issue here:

- https://github.com/dotnet/wpf/issues/6164

It appears the root of the problem stems from using a Chinese application on an OS where English is the default language setting. Specifically, when attempting to render a Chinese character using the Segoe UI font, the anomaly you're observing occurs.

---

问题源于在英文作为默认语言设置的操作系统上使用中文应用程序。具体来说，尝试使用 Segoe UI 字体渲染中文字符时，会出现您所观察到的异常现象。

这个问题还挺难处理的。因为如果判断应用当前的系统语言，会得到英语，然而英语下的默认字体却不是微软雅黑。

walterlv:
After setting the system display language to English, I confirmed through breakpoint verification that the Segoe UI font was indeed applied. The rendering effect is as demonstrated in the image below. It's also worth noting that the Visual Studio interface has adopted the same font rendering effect.

image

love-linger:
* I prefered to use `Microsoft YaHei UI` as the default font for this app on Windows platforms and ignore the system settings.
* Users can change this from `Preferences` menu if they do NOT like the default one.

walterlv:
I agree.

Issue #62, opened 2024-04-08
Suggestion: "Open In Terminal" supports other shells not only git bash


walterlv:
# Interactions

1. Change "Open In Terminal" and "Open In External Tools" from button or drop down to "Split Button"
2. The click of the "Split Button" uses the default shell or default external tool to open
3. The drop down of the "Split Button" list all the supported shells or tools find on this machine

# Functions

I suggest add "pwsh" support to the "Open In Terminal".

This is my pwsh, I like it more than git bash.

![image](https://github.com/sourcegit-scm/sourcegit/assets/9959623/e060d434-b4e6-4859-be4b-47c6dcb3068c)

# Contribution

If you agree with my suggestion, I'd like to fire a new PR to support it.

love-linger:
It sounds great. You can create a PR for this feature. I may need to test it first, or I may need to unify the style of the buttons.

love-linger:
I thought a little bit about this feature.

On macOS and Linux platforms, click on "Open In Terminal" button will open the system built-in terminal, and which shell to use depends on the configuration of this terminal. On Windows, it will open git bash instead because that git provides a built-in bash specifically for Windows.

So I prefer to trait Powershell as an external tool, and try to support it by myself.

walterlv:
> I thought a little bit about this feature.
> 
> 
> 
> On macOS and Linux platforms, click on "Open In Terminal" button will open the system built-in terminal, and which shell to use depends on the configuration of this terminal. On Windows, it will open git bash instead because that git provides a built-in bash specifically for Windows.
> 
> 
> 
> So I <s>prefer to trait Powershell as an external tool, and</s> will try to support it by myself.
> 
> 
> 
> 

Agree. But I prefer that the default terminal can be configured on Windows.

My terminal is the Windows Terminal and the default shell is PowerShell which is profiled for git repositories(See the screenshot above). I know that most PowerShells of Windows users do not support git commands so git-bash exactly should be treat as a default shell on Windows.

I think we can provide a config for the user to change the default terminal or shell on Windows, otherwise, I will always reveal the repo in the file explorer and then open the foler in my Windows Terminal instead of simply click the git bash button.

I'm working on a PR:

1. Keep the open in terminal button as it is now.
2. Change the open in external tools from a button to a split button. Clicking on the button will open the repo using the default extern tool and clicking on the drop down button will open a list of external tools found on this machine.
3. Add two configurations in the settings window, the default terminal/shell and the default external tool.

[...]

PR #64, opened 2024-04-08
Feature: User can set the default terminal/shell + The external tools appears as a SplitButton

walterlv:
Fix: #62 

What does this PR do:

1. Keep the open in terminal button as it is now.
2. Change the open in external tools from a button to a split button. Clicking on the button will open the repo using the default 3. extern tool and clicking on the drop down button will open a list of external tools found on this machine.
Add two configurations in the settings window, the default terminal/shell and the default external tool.

image


walterlv:
Since commit [acbf360](https://github.com/sourcegit-scm/sourcegit/pull/64/commits/acbf3605fd10382490b6082b867ff559813cb969), I've noticed that you have a new suggestion so that I add more commits to implement our new discussions.

love-linger:
I've pushed a similar implementation for this feature.

PR #69, opened 2024-04-12
feat: show git file mode change if exist

### LINKED ISSUE #46: smba 查看 linux 仓库提示很多改动，但是实际上没有
linux 通过 smba 共享文件夹。
然后用 本软件查看，提示很多改动。
我该如何处理让他显示正确。

gadfly3173:
Fix #46  

test branch https://github.com/gadfly3173/gadfly3173/tree/test/file-mode

Screenshots
image

image

image


src/Converters/ObjectConverters.cs
love-linger, line 1: ObjectConverts is a built-in converter in AvaloniaUI.

src/ViewModels/DiffContext.cs
```diff
@@ -3,10 +3,12 @@
 using System.Threading.Tasks;
 
 using Avalonia;
+using Avalonia.Media;
 using Avalonia.Media.Imaging;
 using Avalonia.Threading;
 
 using CommunityToolkit.Mvvm.ComponentModel;
+using SourceGit.Models;
```
love-linger, line 11: Please never add namespace using declaration for SourceGit.*. Instead, use Models.FileModeDiff directly in your code.

gadfly3173: Sorry, I'm new to C#, not familiar with the APIs and codestyle

src/ViewModels/DiffContext.cs
```diff
@@ -66,6 +68,17 @@ public Vector SyncScrollOffset
             set => SetProperty(ref _syncScrollOffset, value);
         }
 
+        public Models.FileModeDiff FileModeDiff
+        {
+            get => _fileModeDiff;
+            set => SetProperty(ref _fileModeDiff, value);
+        }
+
+        public TextTrimming PathTrimming
```
love-linger: PathConverters.TruncateIfTooLong is this way on purpose that is to hide in the middle of the path, but retained 20 characters at the beginning of the path and the entire file name in the end.

gadfly3173: image

I use textTrimming because I want to avoid this

love-linger: Yes. Perhaps I should re-design the toolbar for DiffView.

src/ViewModels/DiffContext.cs
love-linger: You can run `dotnet format` to apply the code-style defined in .editorconfig

love-linger:
I need to redesign the DiffView toolbar. But I will merge the PR first.

PR #82, opened 2024-04-24
Alternative path for save settings

onuch-v:
Some users with not good skills may not found folder with setting when delete installation

Using `~/.source-git` folder can reduce difficult for delete settings.

it's difficult to found cached settings at `~\AppData\Roaming\SourceGit` on Windows for beginning users, when `AppData` is hidden. `SourceGit` is distributed as zip archive and don't have installer which can delete settings at removing installation.

### DIFF
--- a/src/App.axaml.cs
+++ b/src/App.axaml.cs
@@ -43,8 +43,8 @@ namespace SourceGit
                 builder.Append(ex.StackTrace);
 
                 var time = DateTime.Now.ToString("yyyy-MM-dd_HH-mm-ss");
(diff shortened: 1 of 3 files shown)


love-linger:
Since that users have used the `AppData` folder to store the configuration and other caches used by this app, I prefer to add a description of the path used by this app in the README rather than modify it to a new path directly.

onuch-v:
OK. This is the right way

onuch-v:
Add Issue https://github.com/sourcegit-scm/sourcegit/issues/83

PR #107, opened 2024-05-05
Added ability to run program by default for macOS and Linux

dangreene0:
By default the compiled program will not run on macOS or Linux and the user must use `chmod` to enable the ability to execute.

By adding another step in the jobs for macOS and Linux the artifact will now run when the user downloads. 
> c7784b0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,8 @@ jobs:
         run: dotnet build -c Release
       - name: Publish
         run: dotnet publish src/SourceGit.csproj -c Release -o publish -r osx-x64 -p:PublishAot=true -p:PublishTrimmed=true -p:TrimMode=link --self-contained
+      - name: Allow Executing File as Program
+        run: chmod +x publish/Sourcegit
       - name: Upload Artifact
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


dangreene0:
Thanks for the merge!

After testing this, it seems to have worked for Linux but not macOS.

Probably a perms issue. I will possibly look into it and follow up with another PR.

gadfly3173:
`actions/upload-artifact@v4` will package product files into zip, and zip does not support permission management. To solve this problem, you need to package it yourself through tar.
https://github.com/actions/upload-artifact#permission-loss

```yml
- name: 'Tar files'
  run: tar -cvf my_files.tar /path/to/my/directory

- name: 'Upload Artifact'
  uses: actions/upload-artifact@v4
  with:
    name: my-artifact
    path: my_files.tar
```

dangreene0:
> `actions/upload-artifact@v4` will package product files into zip, and zip does not support permission management. To solve this problem, you need to package it yourself through tar. https://github.com/actions/upload-artifact#permission-loss
> 
> ```yaml
> - name: 'Tar files'
>   run: tar -cvf my_files.tar /path/to/my/directory
> 
> - name: 'Upload Artifact'
>   uses: actions/upload-artifact@v4
>   with:
>     name: my-artifact
>     path: my_files.tar
> ```

Thanks for the reply. I found this out through a [Post on StackOverflow](https://stackoverflow.com/questions/72957435/github-actions-macos-app-bundle-as-artifact-is-not-executable-after-download) and I'm working on a PR shortly.

It sure is a strange quirk!

Issue #90, opened 2024-04-27
[BUG] - Can't find repo (Please remove it.)


ennerperez:
In the method FindRepository() `instance.Repositories` returns 0 even when the repo is listed:

![image](https://github.com/sourcegit-scm/sourcegit/assets/1206858/7d7f1f69-13a7-422c-a084-dbd25e68b853)

And it allow to open in file browser and terminal but keep saying:

![image](https://github.com/sourcegit-scm/sourcegit/assets/1206858/24f0ea27-0296-42e1-ae68-8ea80349bf1c)

love-linger:
Yes, it's designed that way. Items in the tree in Welcome page should **NOT** be auto-removed. Instead, it must stay where it should be to notice users that they have a invalid repository.

ennerperez:
But it's not invalid repository, it was not deleted, I have the last release installed but when I run from source code it's detect the previous repo, but won't let me open because of this, so I have to deleted and adding again which is annoying

IDK if this is a bug in prod, or it's only because of the dual execution.

love-linger:
I have reproduced this issue. Here is how I did it:

* Make sure you have closed this app.
* Rename the target repository manually on disk. 
* Open this app, it will check all repositories and remove those not exists on disk.
* Rename the target repository back to it's original name.

It's because `Open In Terminal` and `Open in File Browser` do NOT check repository from `Preference`. Instead, it use the `_id` field directly.

love-linger:
Although it is not clear why your repository was emptied from `Preference`, the 32e685622b5373082d10f35eb13c294671c0ea08 will at least alert the user that there is something wrong with this tree node and they should remove it.

ennerperez:
Still reproducible in 8.10 I just update and again I'm getting the error.

![image](https://github.com/sourcegit-scm/sourcegit/assets/1206858/0f7d84eb-ef22-4ebe-8a5b-ee5beb794679)

ennerperez:
I think is not enough, I found a workaround to my problem: https://github.com/sourcegit-scm/sourcegit/pull/112

PR #112, opened 2024-05-06
FindRepository Fix

ennerperez:
The problem I found in https://github.com/sourcegit-scm/sourcegit/issues/90 was essentially a difference between "Repositories" and "RepositoryNodes" in some scenarios "Repositories" get an empty list when in other hand "RepositoryNodes" get elements inside, that make visible all repos in the UI but not accessible because it's not present in "Repositories"

image

This workaround basically copies existing repos from "RepositoryNodes" into Repositories when the user tries to open it, and for some reason this is not in the "Repository" node from the preferences.

I rather to check this Json file, because as far I saw this, looks like there is no need to have two "Repositories" nodes inside same Json file, but for now this is my workaround.

### DIFF
--- a/src/ViewModels/Preference.cs
+++ b/src/ViewModels/Preference.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
@@ -363,6 +364,20 @@ namespace SourceGit.ViewModels
 
         public static Repository FindRepository(string path)
         {
+
+            if (!_instance.Repositories.Any(i => i.FullPath == path))
+            {
+                var node = _instance.RepositoryNodes.FirstOrDefault(m => m.Id == path);
+                if (node != null)
+                {
+                    _instance.Repositories.Add(new Repository()
+                    {
+                        FullPath = node.Id,
+                        GitDir = Path.GetFullPath(Path.Combine(node.Id, ".git"))
+                    });
+                }
+            }
+            
             foreach (var repo in _instance.Repositories)
             {
                 if (repo.FullPath == path)


love-linger:
I've pushed a commit d87716dc8675ea15218fa401b3f48ed628337bd2 to solve this problem. This issue appear when `Directory.Exists(repo.FullPath)` returns false for some unknown reason on app startup.

love-linger:
Property `Repositories` stores all the managed repositories. Before commit d87716dc8675ea15218fa401b3f48ed628337bd2, it will try to remove those failed with checking `Directory.Exists(repo.FullPath)`.

Property `RepositoryNodes` stores the structure (as a tree) that you group your repositories. It used by the welcome page. For example, here is mine:

image

love-linger:
And one more thing, according to the manual of [Native AOT deployment overview](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=net8plus%2Cwindows#limitations-of-native-aot-deployment), it's not recommanded to use `System.Linq` in this app (and it will introduce a compile warning).

image

ennerperez:
image

Please refer to: https://github.com/sourcegit-scm/sourcegit/pull/114

I'll be removing the reference System.Linq according to the AOT limitations

love-linger:
Make sure that your device is properly mounted to the path '/run/media/xxxx' when this app starts. There is only one line code can cause this problem - `_instance.Repositories.RemoveAll(x => !Directory.Exists(x.FullPath));`

ennerperez:
Yes it's. I'm 100% sure of this, it's my Main work disk, also this happens to me in macOS the Json get malformed sometimes.

The problem it's not deleting the node from Json file, the problem is having duplicates nodes to save the same information for some reason store data in Repository nodes but not in Repository, you can restore this line cuz is fine, it's not the real problem.

ennerperez:
Yes it's. I'm 100% sure of this, it's my Main work disk, also this happens to me in macOS the Json get malformed sometimes.

The problem it's not deleting the node from Json file, the problem is having duplicates nodes to save the same information for some reason store data in Repository nodes but not in Repository, you can restore this line cuz is fine, it's not the real problem.

Please take a look at PR 114.

love-linger:
I've test this pr by `gh pr checkout 112` and the other one by `gh pr checkout 114`. I know exactly what you have changed. As I said before, I do NOT want to merge `Repository` into `RepositoryNode`, because they are different things that store different informations.

PR #114, opened 2024-05-06
Repository Nodes Simplification

ennerperez:


love-linger:
The behaviour with this pr:

image

It's because that the proeprty `Repositories` gethered from `RepositoryNodes` with only top level of nodes. As I said, RepositoryNodes is NOT an item in list, instead, it's an item in a tree!

PR #115, opened 2024-05-06
Close Tab Enhancement

ennerperez:
- Avoid close the app when there is no more tabs open
- Open by default LauncherPage when no tabs detected

### DIFF
--- a/src/ViewModels/Launcher.cs
+++ b/src/ViewModels/Launcher.cs
@@ -1,6 +1,6 @@
 ﻿using System;
 using System.IO;
-
+using System.Linq;
 using Avalonia.Collections;
 
 using CommunityToolkit.Mvvm.ComponentModel;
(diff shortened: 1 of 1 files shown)


love-linger:
Here is write on purpose, see code in `ViewModels/Launcher.cs` at line 106:

```csharp
public void CloseTab(object param)
{
    if (Pages.Count == 1)
    {
        App.Quit();
        return;
    }
     
    // ...
}
```

It's a common behaviour on multi-tabs apps like Chrome.

ennerperez:
Any other example Chrome is not good example, Browser like Opera don-t allow to close all tabs same for Thunderbrid. But this app is not a web brower, is anoying to re open the app when no tabs presents.

love-linger:
As I said, it's feature that write on purpose, not a bug.

PR #121, opened 2024-05-10
Add branch and tag search box to repository Workspace section

bogomolets-owl:
Hello!

This pull request adds two features:
1) Tags display is implemented as a tree to support tag subfolders for organazing tags.
This is done by extending the BranchTree to support tag nodes and modifiyng the xaml markup to tree instead of DataGrid.
2) Ability to search tags and branches in wokspace with a single search box.
The search box implementation is aware of possible large amount of branches/tags and does not spam the search for each input key intentionnaly. The search itself is done via regex and valid regex will work, while invalid will just show a default view.

image

Best regards, Alexander Bogomolets.

### DIFF
--- a/src/Models/BranchTreeNode.cs
+++ b/src/Models/BranchTreeNode.cs
@@ -1,6 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
-
+using System.Text.RegularExpressions;
 using Avalonia.Collections;
 
 namespace SourceGit.Models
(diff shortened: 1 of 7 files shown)


love-linger:
Very good advice. Nice job!!! There is only one point on which I disagree.

**I don't like to display tags as a tree**. 

First, the tag itself is very different from the branch. I think it is only the label of the revision referenced. 

The second is that the tag itself is sorted, see `Commands.QueryTags`, not by name, but by time. This is very important and useful.

love-linger:
I have submitted a similar implementation, but keep the existing structure of tags.

PR #134, opened 2024-05-20
supported accept repository path from cli

seayxu:
supported accept repository path from cli

```ps
sourcegit.exe D:\code\tmp\sourcegit
```

### DIFF
--- a/src/Views/Welcome.axaml.cs
+++ b/src/Views/Welcome.axaml.cs
@@ -8,7 +8,6 @@ using Avalonia.Input;
 using Avalonia.Interactivity;
 using Avalonia.Platform.Storage;
 using Avalonia.Threading;
-using static System.Runtime.InteropServices.JavaScript.JSType;
 
 namespace SourceGit.Views
 {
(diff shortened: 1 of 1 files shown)


love-linger:
1. If you start this app without custom commandline args, it will crash.

image

2. If you have enable `Restore windows` and the last time you shutdown this app with opened repository tabs, there will be no `Welcome` page opened on startup. 

image

And after started, if you open a new tab, it will just replaced to the repository that specified by the commandline. It's very strange.

love-linger:
And it's not recommanded to do this in `Views`. Instead, it should be implemented in `ViewModels`.

seayxu:
Fixed crash without path, to implement this in `View` is fast.

love-linger:

The disadvantage of implementing in `Views.Welcome` is that it will not work until you have a welcome page.

love-linger:
For example, this is the behavior with your PR:

https://github.com/sourcegit-scm/sourcegit/assets/4075673/f991af5f-08f8-4d24-ac2f-0b8aa2c28b70

PR #158, opened 2024-05-31
fix: i18n use en_US.axaml as default language resource

gadfly3173:


love-linger:
If you want to use en_US resource as the fallback for zh_CN, you can simply add these lines into zh_CN.xaml without changing any code.

```diff
 <ResourceDictionary xmlns="https://github.com/avaloniaui" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+   <ResourceDictionary.MergedDictionaries>
+    <ResourceInclude Source="en_US.axaml"/>
+   </ResourceDictionary.MergedDictionaries>
  <x:String x:Key="Text.About" xml:space="preserve">关于软件</x:String>
```

PR #166, opened 2024-06-05
ux: get brush color for LineBackgrounRenderer from themes

ccyr119:


love-linger:
1. In my opinion, the Windows background color provided by this PR is too low in contrast, resulting in the overall gray.
image
2. The new color schema in LineBackgroundRenderer does not fit well with the Light theme
image
The original color schema looks like this
image

ccyr119:
image

love-linger:
I've pushed a commit 62e5ed8a42f871c02b07b23f24de5753fe09014a that allows user overrides default color theme with a json file. For example:

```json
{
   "Color.FG1": "#FF00FF00"
}
```

This file will change the Foreground color for the current active theme.

Issue #199, opened 2024-06-21
[BUG] Crash when staging without selecting any file


FardinHaque70:
![SourceGit Crash](https://github.com/sourcegit-scm/sourcegit/assets/44266051/7b5ffabb-1762-405c-ac8d-f1780093dfee)
SourceGit crashes if i press the single stage button without selecting any file from the unstaged list. 

I'm using the mac version on my M1 Mac.

PR #200, opened 2024-06-21
fix: NRE in WorkingCopy (#199)

gadfly3173:


gadfly3173:
It seems that the Silicon compilation failure is a bug in the .net SDK or github actions?

Issue #175, opened 2024-06-11
[FEATURE] Windows installer and Package management etc. Winget / Choco


tom8zds:
need installer for windows or upload to package management like winget or choco, portable release not convinent when upgrade

lgranie:
I have made a scoop manifest
[sourcegit.json](https://github.com/user-attachments/files/15884120/sourcegit.json)

LouieTian0209:
Agreed. Having a standalone installer for Windows would be a lot better, especially since there are already deb and rpm packages for Linux.

gadfly3173:
I tried to create a manifest for a winget package: https://github.com/gadfly3173/winget-pkgs/commit/d35d48449f92c2bcfcc756e6fa6ad6fcf49839d9 , but encountered some issues:

- Portable package installed via winget is not able to create shortcuts on the Desktop and Start Menu, while open them through the command line throws a `System.DllNotFoundException: Unable to load DLL 'libSkiaSharp' or one of its dependencies`. There is no issue when opening the application directly from the winget installation location or by using the shortcuts located in `%LOCALAPPDATA%\Microsoft\WinGet\Links`.
- winget lacks the ability to automatically detect new release versions, requiring either manually submitting a PR to the `winget-pkgs` repository or using `wingetcreate.exe` to automatically initiate the process and then manually filling out the PR content.

I think a separate installation package is helpful to improve the upgrade experience, which is also the form of installation of most GUI tools in winget.

ZaLiTHkA:
just FYI, searching through the Scoop "apps" list page, I found a user bucket that includes Sourcegit: https://scoop.sh/#/apps?o=false&q=sourcegit

> I have made a scoop manifest [ScoopInstaller/Main#5923](https://github.com/ScoopInstaller/Main/pull/5923)

unfortunately, this PR was closed with a request to reopen it in the "Extras" bucket, but I don't think this has happened yet.

---

not to imply that a Winget package is not required, just a note to say that an installation path is currently available through Scoop.

love-linger:
The description for this app in https://scoop.sh/#/apps?o=false&q=sourcegit is `Native Kubernetes desktop client.`.

I've created a new [package request](https://github.com/ScoopInstaller/Extras/issues/13534) to `ScoopInstaller/Extra` bucket. Hope they can add it to the offical bucket.

ZaLiTHkA:
[...]

PR #202, opened 2024-06-22
feat: add windows x64 installer build with NSIS (#175)

gadfly3173:
To use NSIS generate installer, you need install NSIS from https://nsis.sourceforge.io/Download or run `choco install nsis -y`

NSIS reads the nsi script as ANSI Code Page, so I set the encoding of `build/build.windows.installer.nsi` to ASCII to ensure compatibility.


love-linger:
In my opinion, the key to issue #175 is to provide a more convenient way to self-update rather than requiring an installer.

love-linger:
If someone could help add support for Winget's installation and update methods for this app, this issue would be mostly resolved.

gadfly3173:
According to https://learn.microsoft.com/en-us/windows/package-manager/package/
> Windows Package Manager currently supports installers in the following formats: MSIX, MSI, and EXE.
A installer is required, maybe I should change NSIS installer to MSIX?

PR #204, opened 2024-06-24
fix: show AddFolder menu in context menu of RepositoryNode

gadfly3173:
When displays more than one page of repositories on the welcome page, the new group menu button only a small area on the left side of the bookmark can be called out. This PR is to solve this problem.

### DIFF
--- a/src/ViewModels/Preference.cs
+++ b/src/ViewModels/Preference.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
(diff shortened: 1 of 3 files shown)


love-linger:
I recommanded you just added a button on toolbar to create top-level group (the entry to create sub-group can be found from context menu of desired parent group).

gadfly3173:
The var `parentNode` is due to make the `Create Sub-Group` menu can be showed in the context menu of the RepositoryNode which has group.

https://github.com/sourcegit-scm/sourcegit/assets/28685179/4efb36fe-036a-4e4a-95c9-60a149a0c4b2

love-linger:
> The var `parentNode` is due to make the `Create Sub-Group` menu can be showed in the context menu of the RepositoryNode which has group.

I know what problem you're trying to solve and have thoroughly tested your PR.

In this PR, `Sub-Group` in a repository node has a different meaning than the `Sub-Group` in a folder node. 

Since it's not hard to create a sub-group in a existing parent, we just want an easy way to create top-level group when the tree takes all the vertical space in Welcome page.

love-linger:
I've push a simple implementation to solve this problem

PR #221, opened 2024-06-28
add Color.BadgeFG

ghiboz:
Color.BadgeFG to edit the foreground badge


love-linger:
I'll close this PR, and add this feature manually because the target of branch is not `develop`

Issue #215, opened 2024-06-27
[Feature] Filtrer for tags


Kvnsm:
Wouldn't it be nice to have a tag filter if possible?

![image](https://github.com/sourcegit-scm/sourcegit/assets/17532533/e8f1384a-734e-410a-becf-8fc581d4c3fe)

love-linger:
To try this feature, please download the latest build from https://github.com/sourcegit-scm/sourcegit/actions/runs/9697741836

PR #237, opened 2024-07-04
Make the side panel scroll as a whole

Noisyfox:
This works much better on small screens when repo has lots of tags and branches

side_scroll

### DIFF
--- a/src/Resources/Styles.axaml
+++ b/src/Resources/Styles.axaml
@@ -1205,8 +1205,8 @@
             <Grid>
               <Border Name="PART_Background" CornerRadius="{TemplateBinding CornerRadius}" Background="Transparent"/>
               
-              <Grid Name="PART_Header" ColumnDefinitions="16,*" Margin="{TemplateBinding Level, Mode=OneWay, Converter={StaticResource TreeViewItemLeftMarginConverter}}">
-                <Panel Name="PART_ExpandCollapseChevronContainer">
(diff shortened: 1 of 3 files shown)


love-linger:
If whole panel is scrollable, virtualizing will not work. This can cause stalling when displaying a large number of lists or tree nodes.

love-linger:
for example, this is a user's repo from issue #215, which has 1900+ tags

image

Noisyfox:
Firstly the tags are not shown by default. Secondly they can always filter first, if they really have THAT many tags.

Noisyfox:
My main complatination is even ignore the tags, currently the branch list cannot be properly used on small screens.

Noisyfox:
At the moment, if you have a lot of local branches, you couldn't even see the remotes list nor submodules, not mention tags.
image

Noisyfox:
> > At the moment, if you have a lot of local branches, you couldn't even see the remotes list nor submodules, not mention tags. image
> 
> This is indeed a disaster.
> 
> For a redesign, I would focus on two things
> 
> * Need to support virtualization so that it doesn't get stuck when there are many items
> * Each group should be easily found at any time. For example, if you are scrolling through a list with 1000+ tags, you should also be able to see the grouping of local branches in that state, rather than having to scroll back.

Yeah agreed.

love-linger:
I've pushed a new implementation for this. You can try it with latest develop branch

Noisyfox:
> I've pushed a new implementation for this. You can try it with latest develop branch

That's much better now! Thanks! I can daily drive this instead of SourceTree now~ Fantastic work man!

love-linger:
I've pushed a new commit for this feature. You can try the latest build

PR #252, opened 2024-07-10
ux: Improved 'current branch' visibility

ghiboz:
the current branch is bolder than before, the other are lighter and the current is underlined

image


love-linger:
I've test your PR.

* The built-in monospace font `JetBrains Mono` does not include the `Extra Bold` style.
* I do not like to add a background for TextBlock.

PR #266, opened 2024-07-17
add spaces between chunks

ghiboz:
image


love-linger:
I will not merge this PR for two reasons

* this PR is incompatible with the `Models.TextDiff.ProcessIndicatorForPatch` method, which will used in both hunk mode and user selection mode
* Non-bug fix, non-performance improvement, not improve the convenience of user operation

PR #301, opened 2024-07-30
feature: Allow show uncommitted changes in commits history like TortoiseHG Workspace

workgroupengineering:
image


love-linger:
I'm sorry, I refuse to merge this PR.

* This complicates the `Histories` view, which should show only committed changes.
* This can be confusing. When the current branch changes, uncommitted local changes are also changed to other branches. In my opinion, uncommitted local changes should not be associated with any branch.
* This PR reduces the size of the `CHANGES` view to half its previous size. It shouldn't be. `Changes` view are used more frequently than the `Histories` view.
* I really don't like this UI/UX design

PR #311, opened 2024-08-02
enhance: add CI for arm64

aikawayataro:
Add CI build step for Linux arm64.


aikawayataro:
I've tested it in the aarch64 chroot and it works fine with Xorg. I have not tested it with Wayland, but I believe it should work as well.

PR #353, opened 2024-08-12
Code review suggestions for German translation

NilsPvR:
These are (at least should be) all the suggestions from the code review for #349. They will improve the German translation. 

Thanks to @warappa for the review on #349!


love-linger:
Could you please add translations for these new keys?

```xml
<x:String x:Key="Text.CtrlClickTip" xml:space="preserve">Hold Ctrl to start directly</x:String>
<x:String x:Key="Text.Repository.ShowTagsAsTree" xml:space="preserve">Show Tags as Tree</x:String>
```

PR #373, opened 2024-08-19
Fix missing permissions for OSX app

aikawayataro:
This will fix missing permissions for the OSX application

This also includes a refactoring of the CI to be able to test the package built form pull request and catch this kind of bugs at an early stage

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,68 @@
+name: Build
+on:
+  workflow_call:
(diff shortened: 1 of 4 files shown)


aikawayataro:
It will be necessary to add a new `build.yml` to solution. I've not found a way to do this with dotnet CLI

love-linger:
Tested.
截屏2024-08-20 10 26 27

Issue #338, opened 2024-08-08
[Bug] Sometimes the counter badge will show blank or 0 


gadfly3173:
![1723106254332](https://github.com/user-attachments/assets/09f260d9-682e-4eaa-b844-5e7111e4ebc1)

Usually it appears after I frequently switch between multiple opened repository tabs. It is unable to reproduce stably. It seems to be related to asynchronous execution and https://github.com/sourcegit-scm/sourcegit/issues/286

love-linger:
I can not reproduce this issue. I suggest that the next time you are in this situation, you can use DevTools (press F12 in debug mode) to see if the data is binding correctly.

gadfly3173:
![image](https://github.com/user-attachments/assets/007dbe67-dfaa-417b-b4da-3b7b1ea540ad)
Do you mean check this? It's easier to reproduce in debug mode, maybe because debug mode runs slower.

love-linger:
I'm sorry, I still can not reproduce it with debugging on my PC (i7-14700KF) and mac (M3Max). 

https://github.com/user-attachments/assets/cc54e482-4d29-425a-830a-fd0ec7f4e0ba

Could please you help me debug it?

heartacker:
I have same issue

gadfly3173:
![image](https://github.com/user-attachments/assets/04b3ab38-c44d-40c5-8c15-f8aa07c06ec1)
I tried to print `WorkingCopyChangesCount` when it was called and found that the count obtained was correct, but the UI showed 0.

gadfly3173:
I tried adding `Mode=TwoWay` to the Binding, which did not solve the problem. I also confirmed through debugging that the `WorkingCopy._count property` is not reassigned when switching tabs

gadfly3173:
It seems that Avalonia 11.0.13 does not have this problem. This may be an issue with Avalonia's TextBlock.

love-linger:
I finally reproduced this issue by **switching branches and check the REFS of commit in details view**. `CommitRefsPresenter` has no asynchronous operation, and by outputting its value, I can confirm that the data is correct.

love-linger:
> It seems that Avalonia 11.0.13 does not have this problem. This may be an issue with Avalonia's TextBlock.

This also works for me.

gadfly3173:
I think https://github.com/AvaloniaUI/Avalonia/issues/16456 is related to this issue, and the PR is not released yet.

[...]

PR #383, opened 2024-08-21
fix: invalidate counter visual (#338)

gadfly3173:
Keep the same implementation as `CommitRefsPresenter` to avoid displaying the number of the previous `DataContext`

### DIFF
--- a/src/Views/Repository.axaml.cs
+++ b/src/Views/Repository.axaml.cs
@@ -98,6 +98,7 @@ namespace SourceGit.Views
                 _label = null;
             }
 
+            InvalidateVisual();
             return _label != null ? new Size(_label.Width + 18, 18) : new Size(0, 0);
         }
 
--- a/src/Views/Repository.axaml.cs
+++ b/src/Views/Repository.axaml.cs
@@ -63,11 +63,6 @@ namespace SourceGit.Views
                 FontFamilyProperty,
                 ForegroundProperty,
                 CountProperty);
-
-            AffectsRender<CounterPresenter>(
-                ForegroundProperty,
-                BackgroundProperty,
-                CountProperty);
         }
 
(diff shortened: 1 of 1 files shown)


love-linger:
`CommitRefsPresenter` depends on `DataContext` to draw contents. But `CounterPresenter` only depends on its own property `CountProperty` to draw contents.

```csharp
static CounterPresenter()
{
    AffectsMeasure<CounterPresenter>(
        FontSizeProperty,
        FontFamilyProperty,
        ForegroundProperty,
        CountProperty);

    AffectsRender<CounterPresenter>(
        ForegroundProperty,
        BackgroundProperty,
        CountProperty);
}
```

The issue #338 still exists about this control?

gadfly3173:
> The issue #338 still exists about this control?

The current situation is that after switching repo, the number of the previous repo may be displayed instead of 0. After making this change, I cannot reproduce the problem here.

gadfly3173:
> `CommitRefsPresenter` depends on `DataContext` to draw contents. But `CounterPresenter` only depends on its own property `CountProperty` to draw contents.
> 
> ```cs
> static CounterPresenter()
> {
>     AffectsMeasure<CounterPresenter>(
>         FontSizeProperty,
>         FontFamilyProperty,
>         ForegroundProperty,
>         CountProperty);
> 
>     AffectsRender<CounterPresenter>(
>         ForegroundProperty,
>         BackgroundProperty,
>         CountProperty);
> }
> ```

It seems that `OnDataContextChanged` can be removed.

love-linger:
Since you have added the `InvalidateVisual` in the end of `MeasureOverride`, I suggest you to remove the `CountProperty` from `AffectsRender` call.

gadfly3173:
Confirmed that `AffectsRender` can be all removed.

Issue #398, opened 2024-08-23
Some more ux suggestions


aikawayataro:
Follow-up to #391

* Press ESC to clear search bar and clear focus on it (behavior followed by many apps)

Things to improve:

* Don't focus on the search bar on the home screen after the app just started. (Optional, but I find focused from start search bar a bit attention-grabber)
* Add `/` hotkey for search (Optional)
* When you click the title bar or create bar with the title bar button, the CTRL + F hotkey does not work
* It looks like all empty tabs is the single object - if I enter some text on one of them and open the repository, the tab created after that will show my previous input in the search bar. The same thing happens with just a few empty tabs - they look "synced". Somewhat confusing

love-linger:
> Don't focus on the search bar on the home screen after the app just started. (Optional, but I find focused from start search bar a bit attention-grabber)

Em... I add this feature just because someone wants it works in this way……

> Add `/` hotkey for search (Optional)

Instead we already has a hotkey `Ctrl/Command + F`

> When you click the title bar or create bar with the title bar button, the CTRL + F hotkey does not work

This is a bug

> It looks like all empty tabs is the single object - if I enter some text on one of them and open the repository, the tab created after that will show my previous input in the search bar. The same thing happens with just a few empty tabs - they look "synced". Somewhat confusing

Yes. This page modifies the repository tree that will save to the preference.json. The data must look `synced`.

aikawayataro:
> Em... I add this feature just because someone wants it works in this way……

No, I mean I'd like it to work like this: you open the app - no focus, you open a new tab - focus

> Instead we already has a hotkey Ctrl/Command + F

That's true, but `/` is also a well-known search hotkey

> Yes. This page modifies the repository tree that will save to the preference.json. The data must look synced.

It's very confusing and requires to clean up the search query in a new tab after you open the repository, also you can't keep different queries in different tabs. I think it's better to assign search query to tab and restore it when active tab is changed, this way you can keep synced repository view while keeping query in its tab

love-linger:
> No, I mean I'd like it to work like this: you open the app - no focus, you open a new tab - focus

So, at the first time we open the Welcome page, we do not want to search, but if it is not opened at the first time, we want to search?

BTW, removing the highlighted border can make it better?

> That's true, but `/` is also a well-known search hotkey

I think having `Ctrl/Command + F` is enough for this app

[...]

PR #407, opened 2024-08-25
 Make search filter launcher page specific

aikawayataro:
This will make the search filter tabbed (draft concept)
Just a concept to show how I think #398 could look like

### DIFF
--- a/src/ViewModels/Launcher.cs
+++ b/src/ViewModels/Launcher.cs
@@ -22,8 +22,11 @@ namespace SourceGit.ViewModels
             get => _activePage;
             set
             {
-                if (SetProperty(ref _activePage, value))
+                var old = _activePage;
+                if (SetProperty(ref _activePage, value)) {
                     PopupHost.Active = value;
+                    UpdateSearchFilter(old, value);
+                }
             }
         }
 
@@ -377,6 +380,14 @@ namespace SourceGit.ViewModels
             }
 
             page.Data = null;
+        }
+
+        private static void UpdateSearchFilter(LauncherPage oldPage, LauncherPage newPage) {
+            if (oldPage != null) {
+                oldPage.SearchFilter = Welcome.Instance.SearchFilter;
+            }
+
+            Welcome.Instance.SearchFilter = newPage.SearchFilter;
         }
 
         private LauncherPage _activePage = null;
--- a/src/ViewModels/LauncherPage.cs
+++ b/src/ViewModels/LauncherPage.cs
@@ -24,6 +24,12 @@ namespace SourceGit.ViewModels
             set;
         } = new AvaloniaList<Models.Notification>();
 
+        public string SearchFilter
+        {
+            get => _searchFilter;
+            set => SetProperty(ref _searchFilter, value);
+        }
+
         public LauncherPage()
         {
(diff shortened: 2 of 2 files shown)


love-linger:
For the following reasons, this PR will not be merged into the repository.

* I am skeptical about the necessity and rationality of the requirement of "needing to open multiple `Welcome` pages for managing or querying repositories".
* Since the datasource behind the view is a single piece of data (`ViewModels.Preference.Instance`), even if multiple `Welcome` pages are optimized and some of the data is stored separately, there are still some synchronization behaviours when performing multi-page operations. For example, as mentioned earlier, after expanding a certain group on one `Welcome` page, the same group on all other `Welcome` pages should also be expanded (because the `IsExpanded` state of the group will be saved in `preference.json`, and if the data is inconsistent, it is impossible to determine which one to use).
* The current implementation method makes it easier for users to perceive that the data on the `Welcome` page is actually one piece, which can better guide users on how to use this page.

In the original design, I wanted users to use the page like they use `Ctrl + P` in VSCode, for example:

* `Ctrl+T` to open a new Welcome page
* Enter the partial name of the repository in search box
* Use Up/Down keys to select the repository to be opened
* Press `Enter` key to open the repository

aikawayataro:
> I am skeptical about the necessity and rationality of the requirement of "needing to open multiple Welcome pages for managing or querying repositories".

There is at least one major reason for me: when you open a repository after using search, the new tab (or other Welcome tabs) will contain the same search query. Could be confusing, I have to clear search filter manually this way.

> Since the datasource behind the view is a single piece of data (ViewModels.Preference.Instance), even if multiple Welcome pages are optimized and some of the data is stored separately, there are still some synchronization behaviours when performing multi-page operations. For example, as mentioned earlier, after expanding a certain group on one Welcome page, the same group on all other Welcome pages should also be expanded (because the IsExpanded state of the group will be saved in preference.json, and if the data is inconsistent, it is impossible to determine which one to use).

The draft implementation in this PR doesn't affect this behavior, as it only restores the search filter. The tabs are still "synced" as before, expanded groups are expanded everywhere. There is still a single `Welcome` instance.

> The current implementation method makes it easier for users to perceive that the data on the Welcome page is actually one piece, which can better guide users on how to use this page.

I can't argue with that, but PR doesn't change anything here. It still feels "synced" because nothing has changed from the original behavior.

> In the original design, I wanted users to use the page like they use Ctrl + P in VSCode, for example:

I see. But if you go with tabs, I think used approach brings more confusion.

Thank you for taking a look! Feel free to close this PR. I just wanted to make sure you saw the point I was trying to make :grin:

love-linger:
> There is at least one major reason for me: when you open a repository after using search, the new tab (or other Welcome tabs) will contain the same search query. Could be confusing, I have to clear search filter manually this way.

So what's the meaning if people only search some repository after a `Welcome` page opened, and then open another `Welcome` page if they want to search again?

aikawayataro:
> So what's the meaning if people only search some repository after a Welcome page opened, and then open another Welcome page if they want to search again?

I meant, for example, if I open SourceGit and use search to find the repo, then open a new tab to open a second one, the previous search is still there.
The latest change sort of fixes that, and I think we're done here.
As for the change I suggested, well, there's no practical purpose to having multiple tabs with their search filters. This behavior just seems natural to me.

PR #419, opened 2024-08-27
Fix and add DE localization keys

NilsPvR:
In 184c89e the key `Text.Squash` was updated for all languages but for the German translation the key `Text.CommitCM.Squash`  was edited. I changed this to comply with the other translations.

Other than that I added translations for the recently added keys.

### Thoughts on translations
- `Text.CopyAllText` could have also been translated with "Gesamten Text kopieren" but I liked the idea of the two context menu options "Kopieren" and "Kopiere gesamten Text" both starting with the same word
- `Text.CommitCM.SquashSinceThis` since we used "Vorgänger" instead of "Eltern" to translate "Parent [commit]" I went for "Nachfolger" instead of "Kind" to translate "Child [commit]". Not super happy with the translation but this can always be improved later


NilsPvR:
Added one commit (9337b2b). 
The previous translation stressed the editor: In the editor, only use the monospace font. Whereas the new translation stresses the font: Only use the monosapce font in the editor

PR #429, opened 2024-08-31
feature: Expand Entire File button in diff area

filipeRmlh:


love-linger:
I'm very sorry. Due to the following reasons, this PR will not be merged into the project:

* Using 4000 for `Models.Diff.MaxDiffUnified` is not a good idea to achieve this goal. Maybe we can read the revision file content then combine the diff with it?
* When the file contains a lot of lines, it is hard to find the next diff chunk since we do not provide buttons to goto next or prev diff content or just a mini-map in scrollbar helping user locate it.
* If the file contains too many lines, I am worried about the actual running performance of this feature.
* Currently, this app supports open diff with external tools，and it is very convenient to do this job. Most professional comparison tools default to displaying the entire file content and the changed parts. I think this requirement should be handled by external tools.

PR #436, opened 2024-09-02
enhance: save preference.json instantly

gadfly3173:


love-linger:
Is there any BUG with `preference.json` saving?

gadfly3173:
> Is there any BUG with `preference.json` saving?

Now saving will only be triggered when you manually close the app. If you force shutdown when Windows shutdown, software crashes, etc., `preference.json` will not be saved, which is sometimes a bit confusing.

gadfly3173:
The current PR does not save the configuration when the RepositoryNode changes, which should be safe.

love-linger:
You should solve the problem in this way.

1. Modify `App.Initialize()` in `App.axaml.cs`:
image
2. Modify `Preference.cs` in `ViewModels` by calling `Save()` in `AddNode(...)`, `MoveNode(...)`, `RemoveNode(...)` and `SortByRenamedNode(...)`

gadfly3173:
The reason for using asynchronous operations when not closing the application is that `File.WriteAllText` is a synchronous operation. Frequent calls to it to write files may cause lag in some environments. For example, `WSL2` file operations are performed over the network, and the IO performance is very low.

[9fe8ffc](https://github.com/sourcegit-scm/sourcegit/commit/9fe8ffcea933a74323fcf039c2f6047cb9f945c7) increased the number of saving opertions for the `MoveNode` method. When called synchronously and without deduplication, the `Save` method will be called 3 times in `RemoveNode`, `AddNode`, and `MoveNode` method.

love-linger:
Yes, I know. As I commented in the commit message, using asynchronous operation will reduce the times calling `Save` method, but the issue you metioned above still exists since the real saving operation will take place after 3 seconds or even more if there are another change of `Preference` occurs.

PR #446, opened 2024-09-06
Automatically select next change after (un)stage

hannesbraun:
I think automatically selecting and showing the next change after staging or unstaging is more useful than having nothing selected.

While I think that this is already an improvement, I wasn't able to keep the focus on the (new) selected element so that I could use the arrow keys to further navigate in the changes after (un)staging. I'm still new to Avalonia and might be overlooking something. Maybe you know how to fix this... 😅

### DIFF
--- a/src/ViewModels/WorkingCopy.cs
+++ b/src/ViewModels/WorkingCopy.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 
 using Avalonia.Controls;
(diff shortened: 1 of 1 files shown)


love-linger:
Actually, I don't like this feature.

* As you said, it's hard to auto focus the next change since the background data (local changes) will be re-calculated after file system watcher detects some file in repository changed.
* I don't really want to add too many predictive user actions to this project. When people stage/unstage local changes one-by-one instead of all, which is the _**next**_ they want to check?

love-linger:
I've pushed another implementation for this feature.

hannesbraun:
> I've pushed another implementation for this feature.

Thanks!

> * I don't really want to add too many predictive user actions to this project. When people stage/unstage local changes one-by-one instead of all, which is the _**next**_ they want to check?

Hmm, I wouldn't say it's predictive. It just allows me to further navigate in the changes after a (un)stage with the keyboard. The use case would be (at least for me) if I have a lot of changes and I want to navigate through them with the arrow keys and stage just the ones I want. If the changes are small or very similar this would also allow me to double check the diff just very quick again (by staging each file one by one) before I author the commit. All of this would also be possible using the mouse, however that takes more time. But of course, this use case might not be relevant for anyone. I just thought I'd share it. So I understand if you don't want to add it. It's a pretty good git client already :)

And sorry for my somewhat delayed answer.

Issue #462, opened 2024-09-11
Flatpak for Linux users


lamyergeier:
Flatpak is supported by Gnome a major Linux Open source software developer. It provides easy package management and updating, additionally your software will feature on Gnome Software store. Appimage can't be updated but needs to be manually downloaded. And doesn't features on the Gnome software store.

love-linger:
#93

aikawayataro:
I've got it working with flatpak, but there are problems:

* Signing with gpg is half broken for me on X11 (host gpg-agent fails to spawn pinentry program)
* There's no proper way to spawn host terminal from flatpak. Code changes are required to workaround this

Once these things are fixed, it will be possible to submit the flatpak
I'll report and handle the first one, but the second one needs discussion, will open issue for this

love-linger:
Thanks @aikawayataro. 

Due to some limitations of the `flatpak` mechanism, when adapting to its packaging method, some difficult-to-solve difficulties were encountered. Moreover, some changes will destroy the existing code consistency and increase the difficulty of later maintenance. 

After communicating with @aikawayataro, I decided not to go any further into this issue within the project. Of course, secondary development based on this project is completely possible (this project uses the MIT License).

stone-w4tch3r:
Speaking about termianl issue: [host-spawn] greatly simplifies working with host terminal, and is already built-in into vscode flatpak version. May this help packaging a flatpak?

aikawayataro:
There are also a few highlights on this topic in #479
I don't see such workarounds as a viable solution. What problem will host-spawn solve here?

stone-w4tch3r:
@aikawayataro speaking honestly, from last messages it was not obvious what particular problems you encountered. So I mentioned host-spawn just in case 🙃

Issue #298, opened 2024-07-29
[Feature] Add repositories for apt/rpm?


stone-w4tch3r:
IMHO package repository is better than binary.

I would be glad to implement this feature, but do you think it is needed?

There already exists linux build script, it can be wrapped with [packagecloud publish action](https://blog.packagecloud.io/using-github-actions-ci-to-publish-your-software-or-library-package-to-packagecloud-repository/), for example.

[UPD](https://github.com/sourcegit-scm/sourcegit/issues/298#issuecomment-2509810649)

love-linger:
Thank you very much if you can help. I'm not very familiar with this aspect.

aikawayataro:
After some research I found https://build.opensuse.org/ and it looks like they even provide repositories for Debian, Fedora, OpenSUSE and CentOS which I see as very complete. This might be worth a look.

chidniu:
Considering you are already bulding an .rpm I think it should be relatively simple to provide a Fedora repo via [COPR](https://copr.fedorainfracloud.org/).

AdrienMatricon:
- I think the most natural way to publish the deb packages would be to create an [Ubuntu PPA (Personal Package Archive)](https://help.ubuntu.com/community/PPA), which is basically an apt repository that a user can declare (through `sudo add-apt-repository XXX`) so that the packages it contains are seen by `apt` (you run `apt update` and then you can install them through `sudo apt install XXX` and get the updates with `sudo apt upgrade`)
- PPAs seem to basically be a directory containing the right files and that can be accessed through http, so there are plenty of ways to create one (to get an idea, you can look at all the replies to [this post](https://askubuntu.com/questions/71510/how-do-i-create-a-ppa))
- There is a platform called Launchpad which seems to be the most mainstream way to host PPAs, more info here : https://help.launchpad.net/Packaging/PPA

AdrienMatricon:
Alternately, if publishing sourcegit as a package for various linux distributions seem like a chore, it might be worth looking it into making it a flatpak package hosted on Flathub (those are cross-distribution)

Masgalor:
I am not quiet sure if someone is working on this right now.
So I am just here to say that I am already packaging several projects on OBS for all kinds of distributions. [Vaultwarden](https://build.opensuse.org/project/show/home:Masgalor:Vaultwarden) for example. OBS is the best multi-distro build/publishing platform I know and I can definitely help with that.

[...]

PR #468, opened 2024-09-13
Allow to disable native AOT deployment

aikawayataro:
This will introduce `SourceGitNoAot` variable, which can be set to `true` to disable native AOT deployment.
While AOT could improve application's speed, bundling the dotnet runtime will also increase its size.

The change is useful for Flatpak with the dotnet runtime included, or for a Linux package that will depend on the `dotnet-runtime` package. This is the more idiomatic way for Linux, and will also minimize the amount of disk space used if the system already has the dotnet runtime installed.

This variable is intended to be used by package maintainers, current Linux CI builds should be left as they are.

### DIFF
--- a/src/SourceGit.csproj
+++ b/src/SourceGit.csproj
@@ -20,7 +20,7 @@
     <RepositoryType>Public</RepositoryType>
   </PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


aikawayataro:
Related  #462 #298

PR #473, opened 2024-09-14
Use ArgumentList for commands instead of manual escaping

aikawayataro:
This will refactor command arguments from a single string to an argument array

Argument arrays are easier to work with, and you don't have to worry about escaping arguments. They are also easier to work with because you can split them into multiple lines

I've tested all the features I use, but an extra pair of eyes is really needed :grin:

### DIFF
--- a/src/Commands/Add.cs
+++ b/src/Commands/Add.cs
@@ -12,19 +12,14 @@ namespace SourceGit.Commands
 
             if (changes == null || changes.Count == 0)
             {
-                Args = "add .";
+                Args = ["add", "."];
             }
             else
             {
-                var builder = new StringBuilder();
-                builder.Append("add --");
+                Args.AddRange(["add", "--"]);
+
                 foreach (var c in changes)
-                {
-                    builder.Append(" \"");
-                    builder.Append(c.Path);
-                    builder.Append("\"");
-                }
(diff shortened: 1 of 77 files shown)


love-linger:
BTW, I don't think using 'ArgumentList' is better than 'Arguments' since we have manually escaped the path strings and when creating a specified process, the parameter list is always concated into a single string.

aikawayataro:
Manual escaping is bug-prone and may fail if path contains `"` with the current implementation. 
But the main reason in `ArgumentList` is to provide a more robust and secure way to provide arguments. It is simply easier to work with array than string.

> when creating a specified process, the parameter list is always concated into a single string

https://source.dot.net/#System.Diagnostics.Process/System/Diagnostics/Process.Unix.cs,2cbba2bbbfb82be9

Maybe on Windows. Before this dotnet will parse command to argv anyway, later passed to `execve` inside native interop

In the end, the main goal of this PR is to make commands more readable and easier to maintain

aikawayataro:
I can address the issues, but how do you feel about the change in general? No worries if you don't need this

love-linger:
> I can address the issues, but how do you feel about the change in general? No worries if you don't need this

Both ways are fine for me. I don't have any preference.

PR #479, opened 2024-09-16
Small fixes to support opening terminal in flatpak

aikawayataro:
This also includes a small error message correction, a fix for crash when opening terminal on the Welcome page on Linux, and finally will add support for spawning host's terminal form flatpak.
There's no support to use host's merge tools, as changes will collide with #473 and I still need some time to implement it properly but it is on the way

### DIFF
--- a/src/Native/Linux.cs
+++ b/src/Native/Linux.cs
@@ -78,7 +78,8 @@ namespace SourceGit.Native
             }
 
             var startInfo = new ProcessStartInfo();
-            startInfo.WorkingDirectory = string.IsNullOrEmpty(workdir) ? "~" : workdir;
+            var home = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+            startInfo.WorkingDirectory = string.IsNullOrEmpty(workdir) ? home : workdir;
             startInfo.FileName = OS.ShellOrTerminal;
             Process.Start(startInfo);
         }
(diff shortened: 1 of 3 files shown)


love-linger:
After reading your PR, I have some understanding of how does `flatpak` work. To tell the truth, I start to dislike flatpak - it even requires explicitly calling `flatpak-spawn` when creating an external subprocess...

aikawayataro:
Well, I dislike Flatpak too :grin:. People just keep asking for it
I don't like them mostly because of their size. They are easy to deploy and install, but there are some major drawbacks

> even requires explicitly calling flatpak-spawn when creating an external subprocess...

only if you want to spawn process inside a host, like a terminal or a merge tool, git is fine inside container.

To summarize, I think software like development tools are simply not ready for flatpak. The problems I faced with gpg, it sends xauth data to gpg-agent (well nobody thought gpg will connect to "remote" gpg-agent) and gpg-agent fails to use flatpak's xauth to create a pinentry window with a cryptic error message. It is fixable, but serves a good example of how things are broken inside flatpak.

4deac98c4eb8429d750d5ddfda8fede82145933c alone is still useful though. Can you cherry-pick it or do I need to extract it to a new pull request?

love-linger:
> https://github.com/sourcegit-scm/sourcegit/commit/4deac98c4eb8429d750d5ddfda8fede82145933c alone is still useful though. Can you cherry-pick it or do I need to extract it to a new pull request?

you can create a new pr for that commit, i'll merge it later

PR #483, opened 2024-09-17
Spell checking proposal

GustavoHennig:
**Proposal: Add Spell Checker to CommitMessageTextBox Control**

This pull request introduces a feature proposal to integrate a spell checker into the `CommitMessageTextBox` control. A working proof of concept (PoC) is demonstrated in the attached screenshots (and in the code).

avalonia-spell-checker-demo2  
image

If you think it will be useful to the project, I can go ahead and finish the implementation. I plan to add a new tab in the Preferences window where users can configure language options for the spell checker.

image

There are a few decisions to be made, such as which languages should be embedded by default or whether languages should be downloaded automatically.

The spell-checking library used in this feature is one I recently developed, which I believe will meet the needs of this project.
https://github.com/GustavoHennig/Avalonia.SpellChecker


love-linger:
Thank you for your attention and support to this project.

First of all, for this feature, I prefer the way we integrate `OpenAI`, or the way modern code editor supports multiple languages using external language servers. 

Also, what to do with some words that are not in the dictionary? In some of the editors I have used, if they provide spell checking, they generally provide a way to add or ignore specific words. If this feature is missing, spell checking can be a bad experience in some cases. This is common in the IT industry, where there are a lot of new words and uncommon abbreviations every day.

aikawayataro:
@GustavoHennig why don't you use the system's spell checker? So you don't have to embed anything, languages and dictionaries are handled by the system.

GustavoHennig:
Thank you for your feedback! 

@love-linger I understand the benefits of integrating external language servers, especially in environments like code editors that handle complex syntax and require multi-language support. However, I believe that for basic text spell checking, particularly in an application focused on general text rather than code, a full language server infrastructure might be overkill. I chose Hunspell because it is lightweight, widely adopted in cross-platform applications like Firefox and LibreOffice, and do not require external dependencies.
Regarding the ignore feature, I agree that it’s essential. It's partially implemented, but since this is an MVP, I don’t intend to spend too much time on a library that won’t be heavily used.

@aikawayataro I simply chose the easiest way to solve the problem, and I'm not sure if Linux and macOS have native spell checkers, I know for sure that it would greatly increase the complexity. Perhaps I could create abstractions to support them in the future. What I do know is that Hunspell is used by many cross-platform apps, and I believe even Chrome and macOS use it internally. I need to investigate this further.

I've always used SourceTree, and I’m glad this project was created. I finally found an opportunity to add a feature I've always missed in SourceTree: a multilingual spell checker. SourceTree only uses the OS spell checker, and there’s no way to configure multiple languages (which is my particular use case).

Anyway, I’m willing to spend some time improving this, but depending on the complexity, I’m afraid the benefits might not be worth the effort. I may end up simply maintaining some sort of SourceGit extension.

ameaninglessname:
It also has the `add to dictionary` option in the context menu.

PR #526, opened 2024-09-29
fix: avoid NRE in ContextMenuExtension

gadfly3173:
Sometimes the context menu may be set to `null` before opening. For example, the `Avalonia` repository has many branches. If you right-click multiple times before the context menu of the branch is opened, NRE will appear.

### DIFF
--- a/src/Views/ContextMenuExtension.cs
+++ b/src/Views/ContextMenuExtension.cs
@@ -1,6 +1,5 @@
 ﻿using System.ComponentModel;
 using Avalonia.Controls;
-using Avalonia.Threading;
 
 namespace SourceGit.Views
 {
@@ -15,10 +14,7 @@ namespace SourceGit.Views
             menu.Closing += OnContextMenuClosing; // Clear context menu because it is dynamic.
 
             control.ContextMenu = menu;
-            Dispatcher.UIThread.InvokeAsync(() =>
-            {
-                control.ContextMenu?.Open();
-            });
+            control.ContextMenu?.Open();
         }
 
         private static void OnContextMenuClosing(object sender, CancelEventArgs e)
--- a/src/Views/ContextMenuExtension.cs
+++ b/src/Views/ContextMenuExtension.cs
@@ -1,6 +1,6 @@
 ﻿using System.ComponentModel;
-
 using Avalonia.Controls;
+using Avalonia.Threading;
 
 namespace SourceGit.Views
 {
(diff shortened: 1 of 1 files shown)


love-linger:
In fact, `ContextMenuExtension.OpenContextMenu` is being called in UI thread, see:

image
image

and the code of `ContextMenu.Open` is
```csharp
        private void Open(Control control, Control placementTarget, PlacementMode placement)
        {
            if (IsOpen)
            {
                return;
            }

            if (_popup == null)
            {
                _popup = new Popup
                {
                    IsLightDismissEnabled = true,
                    OverlayDismissEventPassThrough = true,
                };

                _popup.Opened += PopupOpened;
                _popup.Closed += PopupClosed;
                _popup.Closing += PopupClosing;
                _popup.KeyUp += PopupKeyUp;
            }

            _popup.SetPopupParent(control);

            _popup.Placement = placement;

            //Position of the line below is really important. 
            //All styles are being applied only when control has logical parent.
            //Line below will add ContextMenu as child to the Popup and this will trigger styles and they would be applied.
            //If you will move line below somewhere else it may cause that ContextMenu will behave differently from what you are expecting.
            _popup.Child = this;
            _popup.PlacementTarget = placementTarget;
            _popup.HorizontalOffset = HorizontalOffset;
            _popup.VerticalOffset = VerticalOffset;
            _popup.PlacementAnchor = PlacementAnchor;
            _popup.PlacementConstraintAdjustment = PlacementConstraintAdjustment;
            _popup.PlacementGravity = PlacementGravity;
            _popup.PlacementRect = PlacementRect;
            _popup.CustomPopupPlacementCallback = CustomPopupPlacementCallback;
            _popup.WindowManagerAddShadowHint = WindowManagerAddShadowHint;
            IsOpen = true;
            _popup.IsOpen = true;

            RaiseEvent(new RoutedEventArgs
            {
                RoutedEvent = OpenedEvent,
                Source = this,
            });
        }
```

I think the problem may caused by these code of `SourceGit`
```csharp
private static void OnContextMenuClosing(object sender, CancelEventArgs e)
{
    if (sender is ContextMenu menu && menu.PlacementTarget != null)
        menu.PlacementTarget.ContextMenu = null;
}
```

love-linger:
BTW, this code is introduce by 0dea7ed0e2fa5b30d2d87282bcc2d7dd4a70815b with issue #140

gadfly3173:
Maybe just with `control.ContextMenu?.Open();` is fine, or someone has a better solution?

love-linger:
> Maybe just with `control.ContextMenu?.Open();` is fine, or someone has a better solution?

I can not reproduce this issue, but I think you are right.

PR #543, opened 2024-10-04
new feature  about push new branch directly

EastRedBlack:
push new branch to remote ,just click new branch checkbox,
Due to work requirements, there are many times when we cannot directly push to the main repository, but we still need a pull request. Therefore, the best approach is to push directly to a new remote branch and then merge it. Although this practice is not very common, tools like  (TortoiseGit) have this functionality and can be referenced.


love-linger:
Why not create a new branch locally first and then push it directly to the remote? Or use a workflow like `git-flow`?

image

> [!NOTE]
> When the selected local branch has no tracking upstream, and there's not a branch with the same name of selected local branch, a branch in `${SELECTED_LOCAL_BRANCH_NAME} (new)` will be automatically selected - it means a new remote branch with the same name of selected local branch will be created and push current selected local branch into it.

EastRedBlack:
Based on  TortoiseGit Function 
image
Simplifying the push process, having too many local branches requires self-management.

EastRedBlack:
Ok. This is just my suggestion.

PR #577, opened 2024-10-19
Fix untracked files being staged when hidden

navad:
Current version of SourceGit contains an optimization to perform a `git add <repo path>` operation when the number of files being staged equals the number of changes. While this optimization makes perfect sense, it fails to accommodate for untracked files, if those are marked **hidden** in the UI ("Include Untracked Files" is disabled)

Before staging: 
image

After staging:
image

This PR includes a fix to check and make sure there are no untracked files present in the repo before performing this operation in order to avoid adding those unwanted files.

### DIFF
--- a/src/Commands/QueryHasUntrackedFiles.cs
+++ b/src/Commands/QueryHasUntrackedFiles.cs
@@ -1,0 +1,24 @@
+namespace SourceGit.Commands
+{
+    public partial class QueryHasUntrackedFiles : Command
+    {
+        private bool _hasUntracked = false;
+
+        public QueryHasUntrackedFiles(string repo) {
+            WorkingDirectory = repo;
+            Context = repo;
+            Args = "ls-files --others --exclude-standard";
+        }
+
+        public bool Result()
+        {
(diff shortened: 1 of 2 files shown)


love-linger:
I've pushed my implementation of this feature. It just using `-u` option of `git add` command to ignore untracked files.

navad:
Thanks for applying a fix! 
`git add -u ...` will indeed filter out untracked files but might implicitly stage other files, that's why I preferred a more explicit approach in the PR when there are any filtered-out files in the UI

love-linger:
> Current version of SourceGit contains an optimization to perform a git add <repo path> operation when the number of files being staged equals the number of changes

Why I perfer to use `git add [-u] .` instead of `git add -- <pathspec> ...`? Here is the reason:

The `--pathspec-from-file=<file>` parameter of `git add` command is added since `2.25.0`, which is not the minimal version of git supported by SourceGit. So, we can only use `-- <pathspec> ...` format. When there are many of file to stage, the argument list can be too long, and may fail on some platform.

PR #599, opened 2024-10-24
refactor: improve pt_BR localization clarity and consistency

dougcunha:
I revised several terms that sounded unusual in Portuguese. Since certain GIT commands are typically not translated by the developer community in Brazil, and some other terms have optional translations, I aimed for a more balanced approach. Additionally, I took the opportunity to translate features that hadn't been localized yet.

Please feel free to disregard this pull request if the translation approach doesn't align with your preferences.


love-linger:
* We should not bind UI display language with commit message. These are two completely different things. Especially in some open source software projects (there may be developers from different countries involved)
* Since we have added the ability to customize AI prompts, it is not difficult for users to achieve this goal by configuring it themselves.

For the above reasons, I have manually picked commit 63f75dc589e6c6effee498222b247d2616c5b7d0, and the commit c6d308be2c526d817f962c5545a2a423817cfd4f will not be merged into this repository

dougcunha:
I must admit, I didn't quite understand your point. I assume SourgeGit is not exclusively for open-source projects, correct? In my case, I work on projects from various countries and need to generate commit messages in each local language. Setting up the prompt every time I generate a commit for a repository is not very practical, which is why I thought of a feature that would provide a language selection menu when generating commit messages.

In any case, I'll maintain this feature in a branch within my fork and prepare a custom version of SourceGit that includes it.

love-linger:
Sorry, I have too many things to do, so I didn't carefully review the code. I regard the new language option as the ui display language configuration. You can create a new pr for another commit.

dougcunha:
No problem! It’s possible that the way I implemented this feature has made the UI a bit unclear. The goal is simply to have a quicker method to include an instruction for generating the commit message in the selected language, without needing to edit OpenAI prompts directly in the preferences window.

Ideally, this setting could be saved individually for each repository, although that would add complexity. I’m open to any suggestions you might have for improvement.

love-linger:
> Ideally, this setting could be saved individually for each repository, although that would add complexity.

How about changing the way to configure OpenAI prompts & response language (OpenAI profiles) similar to the way we configure commit template and then shows all profiles when user click the `User OpenAI to generate commit message` button?

love-linger:
After the release of version `8.36`, I will submit a new solution of integrating OpenAI services.

PR #600, opened 2024-10-25
ci: add Localization Check CI

gadfly3173:
Add localization check CI and translations missing keys in documents. Using `nodejs` but not `python` is my personal preference. Maybe I can do this with `.Net` unit tests, but I'm not very familiar with how to write this in `.Net`.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -292,3 +292,8 @@ indent_size = 2
 end_of_line = lf
 [*.{cmd,bat}]
 end_of_line = crlf
+
+# YAML files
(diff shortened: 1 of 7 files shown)


gadfly3173:
You can check the execution behavior at https://github.com/gadfly3173/sourcegit/tree/develop

love-linger:
> Maybe I can do this with `.Net` unit tests, but I'm not very familiar with how to write this in `.Net`.

This implementation is good enough. There is no need to use `.NET` for it.

PR #601, opened 2024-10-25
doc: Make the badges clickable and redirect to the TRANSLATION.md

gadfly3173:


love-linger:
A small suggestion: should we move `localizaton-check.js` to `build/scripts` directory? This folder contains all resources and scripts those used by `Github Action`. 

The advantage of it is that on Unix-like platforms, `.github` is not visible by default.

gadfly3173:
I have no particular preference for the file location. I put it in `.github` just because `build` is ignored by `.gitignore` and has little to do with other release scripts used.

PR #605, opened 2024-10-27
feat: support language for commit message generation

dougcunha:
- Extend the GenerateCommitMessage constructor to include a language parameter and use it in the selected language prompt logic.
- Add property for commit message language with default value "English".
- Add integration related configuration strings for commit message language.
- Add property for CommitMessageLanguage in RepositoryConfigure.
- Add language parameter to AIAssistant constructor and propagate it to GenerateCommitMessage.
- Add integration configuration options in the repository settings view.

SourceGit_e1QzjvGHnn

@love-linger I’d like to hear your thoughts on this approach. Please let me know if you’ve considered any alternatives. I also just noticed your comment on the other pull request.

### DIFF
--- a/src/Models/NumericSort.cs
+++ b/src/Models/NumericSort.cs
@@ -50,22 +50,9 @@
                 string sub2 = new string(tmp2, 0, loc2);
                 int result;
                 if (isDigit1 && isDigit2)
-                {
-                    // compare numeric values
-                    if (sub1.Length == sub2.Length)
-                    {
-                        // if length is the same, lexicographical comparison is good also for numbers
-                        result = string.CompareOrdinal(sub1, sub2);
-                    }
-                    else
-                    {
-                        result = sub1.Length.CompareTo(sub2.Length);
-                    }
-                }
+                    result = loc1 == loc2 ? string.CompareOrdinal(sub1, sub2) : loc1 - loc2;
                 else
-                {
-                    result = string.Compare(sub1, sub2, System.StringComparison.Ordinal);
-                }
+                    result = string.CompareOrdinal(sub1, sub2);
 
                 if (result != 0)
                     return result;


love-linger:
I'm very sorry, I'm rewriting the integration method of OpenAI. I will reply under this PR after my modifications are completed. At that time, you can test and make new modifications.

love-linger:
Now, we support to configure multiple OpenAI services.

image

* If there's only one service available, clicking the `Use OpenAI to generate commit message` button will directly generate commit message by it.
* If there're multiple services available, clicking the `Use OpenAI to generate commit message` button will show a context menu to select
image

love-linger:
So, you can configure multiple services (for example, one using English and another using Português) and the select the one you want when generate commit message in different repositories.

dougcunha:
I think your solution looks great. Perhaps an improvement could be having a per-repository setting for which pre-configured AI service I want to use. If it's set, the context menu wouldn't open, generating the message directly.

I'll sync your changes and test it on my end as well.

love-linger:
> I think your solution looks great. Perhaps an improvement could be having a per-repository setting for which pre-configured AI service I want to use. If it's set, the context menu wouldn't open, generating the message directly.
> 
> I'll sync your changes and test it on my end as well.

At present, since there are many AI models to choose from, and they are updated relatively quickly, I am testing different models myself, such as: some are better at understanding Chinese, and some are more inclined to understand source code. I personally prefer to keep the context menu for multiple services (multiple prompts), so that there is more choice, or the opportunity to compare multiple outputs and adopt the best results.

dougcunha:
> At present, since there are many AI models to choose from, and they are updated relatively quickly, I am testing different models myself, such as: some are better at understanding Chinese, and some are more inclined to understand source code. I personally prefer to keep the context menu for multiple services (multiple prompts), so that there is more choice, or the opportunity to compare multiple outputs and adopt the best results.

I understand your point of view, and the idea of comparing results from different models before choosing the best is interesting. While this might not be as necessary when using paid models like GPT-4 or Claude 3.5, it makes more sense for those using local models like Ollama.  However, it would still be beneficial to have an optional configuration per repository for those who don't need to test multiple models.

love-linger:
Thanks. I agree with you that having a per-repository setting for prefered AI service. I will try to add this feature myself tomorrow.

dougcunha:
Thanks leo.
Let me know if you need any help.

Em seg., 28 de out. de 2024 às 09:46, leo ***@***.***>
escreveu:

love-linger:
Done.

image

PR #673, opened 2024-11-10
Add children list to the commit base info view

czarkoff:
Useful for navigation between the commits.

The implementation is suboptimal.  It would probably make more sense to search for children when the view is displayed, but that would require much more substantial changes as currently commit list is not available in the view.

That said, in my testing this implementation appears to perform acceptably.

### DIFF
--- a/src/Commands/QueryCommitChildren.cs
+++ b/src/Commands/QueryCommitChildren.cs
@@ -1,0 +1,34 @@
+﻿namespace SourceGit.Commands
+{
+    public class QueryCommitChildren : Command
+    {
+        public QueryCommitChildren(string repo, string sha)
+        {
+            WorkingDirectory = repo;
+            Context = repo;
+            _sha = sha;
+            Args = $"rev-list --children --all {sha}^..";
+        }
+
+        public string[] Result()
+        {
+            var rs = ReadToEnd();
+            if (!rs.IsSuccess)
+                return [];
+
+            int start = rs.StdOut.IndexOf($"\n{_sha}");
+            if (start != -1)
+            {
+                int end = rs.StdOut.IndexOf('\n', start + 1);
+                if (end == -1)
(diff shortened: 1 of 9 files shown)


czarkoff:
FWIW to me it would make sense to change parent and children lists from `List<string>` to `List<Commit>`. That would also allow for showing tooltips with useful info in the details view.

love-linger:
To be honest, I don't really like this feature - especially the design of `Children`.

> FWIW to me it would make sense to change parent and children lists from List<string> to List<Commit>. That would also allow for showing tooltips with useful info in the details view.

I also don't agree with this modification. Relevant, if necessary, I will provide a custom text block control that used to display `Commit.SHA`. Only when its content changes will it asynchronously obtain the information needed for tooltips.

czarkoff:
You don't find this feature useful in principle, or you would prefer a different implementation?  An alternative would be to search for the commit's children when it is being displayed.  Or maybe you have some other idea?

czarkoff:
I've reverted the previous attempt and made another one. This time children are fetched asynchronously for a specific commit while the view is being rendered.  If I understood your feedback correctly, that should be more in line with your approach.

Now, it is possible that I didn't understand your feedback correctly.  If that is the case,  I would appreciate if you could explain what kind of implementation for this feature you would like to see.  Or close the PR if you don't regard this functionality as a desirable feature.

Either way I hope I didn't waste too much of your time unnecessarily,  and I am really sorry if I did.

love-linger:
Nice job. It's better now.

love-linger:
I'm sorry. I will revert this PR due to the performance of `Commands.QueryCommitChildren`. It takes more than 2 seconds on my PC (AMD 7950X).

czarkoff:
I see.  It was much faster with repos I tried, so thank you for the pointer at a better test. I will look into this next weekend.

PR #687, opened 2024-11-12
Add missing de_DE keys

NilsPvR:
Most were straightforward. I translated `Text.Hotkeys.Repo.CreateBranchOnCommit` a bit more freely to reduce the amount of characters.


love-linger:
It's not necessary to include the commit that created by github-actions[bot].

PR #700, opened 2024-11-14
update translation for russian

AquariusStar:


AquariusStar:
I have some problems with my PC. 
Therefore, I see a strange error status.

PR #701, opened 2024-11-14
fix: commit histories double click error

albertodlc:
**Error**
When a commit was selected (single click without checkout) and the user double tapped in an empty part of the histories ListBox
image

The Checkout message was displayed for the last selected commit. 
image

**Fix**
It shouldn't execute that action unless the user double tap on a commit. I validated the name of the pressed UI element and ignore the empty part of the checkbox.


love-linger:
I've pushed my implementation for this issue. See commit 4835c3f1e9c5b1fb89e5fc5bc39df172a8a43a9c

PR #708, opened 2024-11-16
File histories from commit view

czarkoff:
Currently when the file history is requested,  `git log` runs against the current branch.  Thus if file history is requested for a file from a commit from another visible branch,  the selected commit does not show up in the history.  When the file was added in another branch and was not yet seen in the current branch,  the history view is just empty.

This PR makes the log run against the selected commit,  which makes most sense to me.

Another approach would probably be to find a first parent reference containing current commit and run log against that,  though to me it would be a bit counter-intuitive.


NilsPvR:
Hi, make sure your target branch for the PR is develop and not master.

PR #710, opened 2024-11-17
Show the list of children in the commit details

czarkoff:
This is an updated version of a previously merged and reverted #673.

The general case remains almost as slow as it was, with the following optimizations:
* The tree is walked one commit less for each ref which contains the commit.
* Git is doing slightly less work as the search is based on parents,  not on children.  As the output is filtered by us anyway, this should be more efficient,  though I didn't measure to be frank.

Additionally,  the git command now respects:
* Filter settings.  The refs that are not visible are not searched, which should eliminate a certain amount of commits for git to chew through.  The impact should not really be worthwhile,  as either most of the time is spent on the references that don't contain the searched commit,  or the searched commit is far enough from the refs for git to take time to reach it anyway.
* `Preference.Instance.MaxHistoryCommits`.  This might provide a positive impact in cases when not all of the common ancestors made it to the graph,  which in practice should not be the case.

I don't see any further optimization opportunities.

That said,  I've tried to address the performance problem by loading children asynchronously.  In the repos that I commonly use this produces no effect whatsoever as the children are found before the view is actually drawn.  I don't have access to the `UnrealEngine` repo,  so I used Torvalds' Linux tree as my test stand,  and indeed for the commits that are further up the tree the commit details get drawn first without children,  and then shortly children blink into the view;  the further I go down the tree,  the longer it takes for the commit details to get rendered,  so that at some point children tasks starts producing results before the view is rendered.  In my testing I only saw times above 0.5 seconds when GC was running;  though performance issues are IO-bound in this case,  so I am not sure whether this figure is representative.

**TL;DR**:  in my testing this iteration is usable with big repos.

### DIFF
--- a/src/Commands/QueryCommitChildren.cs
+++ b/src/Commands/QueryCommitChildren.cs
@@ -5,12 +5,14 @@ namespace SourceGit.Commands
 {
     public class QueryCommitChildren : Command
     {
-        public QueryCommitChildren(string repo, string commit)
+        public QueryCommitChildren(string repo, string commit, string filters)
         {
             WorkingDirectory = repo;
             Context = repo;
             _commit = commit;
-            Args = $"rev-list --parents --all ^{commit}";
+            if (string.IsNullOrEmpty(filters))
+                filters = "--all";
+            Args = $"rev-list --parents {filters} ^{commit}";
         }
 
         public IEnumerable<string> Result()
--- a/src/Commands/QueryCommitChildren.cs
+++ b/src/Commands/QueryCommitChildren.cs
@@ -15,19 +15,19 @@ namespace SourceGit.Commands
             Args = $"rev-list --parents {filters} ^{commit}";
         }
 
+        protected override void OnReadline(string line)
+        {
+            if (line.Contains(_commit))
+                _lines.Add(line);
+        }
+
         public IEnumerable<string> Result()
         {
-            var rs = ReadToEnd();
-            if (!rs.IsSuccess)
-                yield break;
-
-            foreach (string s in rs.StdOut.Split('\n', StringSplitOptions.None))
-            {
-                if (s.Contains(_commit))
-                    yield return s.Substring(0, 40);
-            }
+            Exec();
+            return _lines;
         }
 
         private string _commit;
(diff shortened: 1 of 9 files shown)


czarkoff:
@love-linger,  just wanted to ask for your feedback.  Is this still not good enough in your opinion?  Or do you have some other concerns?

love-linger:
I am currently testing this PR. And is reviewing the GIT documentation to see if there is a more efficient way.

Here, I need to clarify that I have never seen this feature in any other Git clients. I have not encountered corresponding requirements in my own use.

love-linger:
* I really didn't expect this to happen.
截图20241119160446

* I recommanded you using `Filters` and `Check refs that contains this commit` instead.

czarkoff:
To the second point:  Filters are already used.  I considered doing `Check refs that contains this commit`, but it would walk the tree another time,  so the end result will be slower.  I can give it a shot though.

I'll check the first point,  I didn't see such things happening in my testing.

love-linger:
If you really need this feature, I suggest you implement it in another way since it is not commonly used. 

For example: like `Check refs that contains this commit` function, add a button after the commit's SHA.
When you click it, a popup will shows the direct children of that commit.

czarkoff:
You see,  the point is to be able to navigate relatively quickly up and down the branch.  This is something I normally do several times a day,  and it was the reason I've used `gitk`,  and is the reason why I still use it despite this project being so incredibly better in so many ways.

Would you accept this change with either repo or project configuration option to show the children?  If the option is unset,  the command is not run, the list is not populated,  and no UIThread callback happens.  The view would not need any changes compared to what was already done,  the widgets in the grid raw would just remain constantly hidden.

P.S.:  I've checked the image,  and either I don't understand your point or everything is actually fine.  This commit just has all those children.

love-linger:
I have merged this PR and made some modifications:

* Since the entire commit details panel already uses `ScrollViewer`, `WrapPannel` is used to display CHILDREN, as shown in the figure:
image
* Modified `Commands.QueryCommitChildren` based on the code style in the project
* Other UI/UX modifications

PR #719, opened 2024-11-19
enhance: introduce template engine for commit templates

### LINKED ISSUE #704: [Feature] Advanced Template for commits
In commit templates allow to filter branch name and file list by regex.

Proposed syntax
`${regex_T:{<filter regex>}:{<replacement regex>}} where T is "b" for branch, "f" for files.`

Example:
`${regex_b:{(.*)\/(issue-\d+).*}:{ID: $2 Type:$1}}`
to convert branch name "feature/issue-1234-add-new-feature"
into: "ID: issue-1234 Type:feature"

aikawayataro:
This replaces the regex-based approach to templates with a simple non-failing and forgiving parser.

Now there are 3 types of variables:

* Regular variable, it just expands to its value, e.g. `${FILES}` (`BRANCH`, `FILES_COUNT`, `FILES` are supported)
* Sliced variable, it expands to special sliced value limited by `N`, e.g. `${FILES:N}` (currently only `FILES` is supported)
* Regex variable, same as regular variable, but before expanding it replaces `regex` with `replacement`, e.g. `${BRANCH/regex/replacement}`. Supports all variables that regular variables support. No newlines allowed inside `regex` and `replacement` because that's error prone. Capture groups (`$1`, `$2`, ...) are supported, as expanding directly uses `Regex.Replace(String, String)`.

To escape variables within text, use `\` before `$`, e.g. `\${FILES}`. It is also used to escape `/` within `regex` and `}` within `replacement` for regex variables. `\` for escaping is only valid in those contexts, otherwise it behaves like a regular character, e.g. `C:\a\b` => `C:\a\b`, `\\` => `\`.

I also changed the spelling of variables to be more recognizable and "standard", but legacy spellings like `files`, `branch_name` were kept for compatibility.

Adding a variable is trivial, add an entry to `s_variables` for regular ones and/or to `s_slicedVariables` for sliced ones.
Unknown variables are silently expanded to an empty string, but I propose to add some warning logic for this.

Since there is no testing framework used for SourceGit, I didn't add a test for the template engine, but I tested it externally and it works fine. I also did some manual testing inside SourceGit, but final check is needed.

Close #704

### DIFF
--- a/src/Models/CommitTemplate.cs
+++ b/src/Models/CommitTemplate.cs
@@ -1,7 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-using System.Text.RegularExpressions;
+﻿using System.Collections.Generic;
 
 using CommunityToolkit.Mvvm.ComponentModel;
 
@@ -9,9 +6,6 @@ namespace SourceGit.Models
 {
     public partial class CommitTemplate : ObservableObject
     {
-        [GeneratedRegex(@"\$\{files(\:\d+)?\}")]
-        private static partial Regex REG_COMMIT_TEMPLATE_FILES();
-
         public string Name
         {
             get => _name;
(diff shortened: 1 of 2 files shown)


aikawayataro:
If this is merged, any thoughts on highlighting for the commit template editor?

love-linger:
Since, `TemplateEngine` is not related to `git` commands and only deal with datas, I suggest you move it to the `Models` directory.

PS: it is better to run `dotnet format` once time.

aikawayataro:
Ok will move it to `Models`. BTW I ran dotnet format on this, is there something wrong?

love-linger:
image

Sometimes blank lines are used between functions, and sometimes they are not.

aikawayataro:
Oh, I group them by what they do this way. The `dotnet format` didn't touch this, so I thought it was normal. Not a frequent csharp programmer, will fix that :grin:

love-linger:
You can use 

```csharp
#region SOME_DESC
#endregion
```

to group your codes.

PR #734, opened 2024-11-23
feat: context menu for a commit in commit message

czarkoff:
When a commit message happens to contain a commit link and the user elects to right-click it, instead of navigating them to the commit,  present a menu with options to navigate to it or to copy SHA.

I've reused translations for "Copy SHA." "Navigate to" remains untranslated in most locales,  and the English wording choice might be questionable.

### DIFF
--- a/src/Resources/Locales/de_DE.axaml
+++ b/src/Resources/Locales/de_DE.axaml
@@ -597,6 +597,7 @@
   <x:String x:Key="Text.SelfUpdate.IgnoreThisVersion" xml:space="preserve">Diese Version überspringen</x:String>
   <x:String x:Key="Text.SelfUpdate.Title" xml:space="preserve">Software Update</x:String>
   <x:String x:Key="Text.SelfUpdate.UpToDate" xml:space="preserve">Es sind momentan kein Updates verfügbar.</x:String>
+  <x:String x:Key="Text.SHALinkCM.CopySHA" xml:space="preserve">SHA kopieren</x:String>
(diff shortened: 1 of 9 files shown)


love-linger:
Why not click the SHA, navigate to the target COMMIT then copy the SHA if it is the right one you want to copy after checking?

czarkoff:
> Why not click the SHA, navigate to the target COMMIT then copy the SHA if it is the right one you want to copy after checking?

Well,  navigation cancels current selection,  and you might need quite a bit of it in some cases.  For instance I was trying to select some of the commits for cherry-pick from a list of cherry-picked commits elsewhere:

image

The ability to do so without navigation came in very handy.

czarkoff:
Why have you removed the tooltips for the commits?  Are there some negative side effects that I didn't recognize?

love-linger:
> Why have you removed the tooltips for the commits? Are there some negative side effects that I didn't recognize?

4 reasons:

* Run command in UI thread is not recommended
* Like parent SHA, can we cache the queried commit to avoid fetching it if we hover it again?
* As I said, you haven't provided a proper DataTemplate for the tooltip
* I'll provide a common control for all SHA

czarkoff:
Understood,  thank you for explanation.  Please let me know if you wish me to do some of that.

ChiahongHong:
@czarkoff  @love-linger
There might be a bug that needs to be fixed:

https://github.com/user-attachments/assets/5c385405-b5b0-46b1-bbf1-7ee7a74bae97

love-linger:
Fixed by PR #741

PR #760, opened 2024-11-27
[Feat] New Context Menu - Copy commit info to clipboard

albertodlc:
*With branches and tags*
image

*Only commit info*
image


love-linger:
Most of this information can be copied from the details panel. The right-click menu in the list is mainly used to generate information like `HASH - SUBJECT`, or to generate multiple informations of selected commits.

albertodlc:
In the details panel the Refs/Tags cannot be copied. Neither all the information in a quick way.
Also I think is more intuitive having the copy options on the right click menu too, like most programs have.

Both copy mechanism (detail panel and right click menu) can coexist.

albertodlc:
@love-linger is there any possibility of merging this? If not I would prefer to close it.

love-linger:
I'm sorry. I do NOT think this PR is necessary to be merged.

PR #785, opened 2024-12-04
Option to delete only not merged branches

ltetak:
As a user, I would like to ensure I don't delete not merged branches accidentally. This is also helpful when cleaning up local branches.
On additional change would be helpful here to add --prune in the fetch option. I can add that separately.

### DIFF
--- a/src/Commands/Branch.cs
+++ b/src/Commands/Branch.cs
@@ -34,12 +34,20 @@
             return cmd.Exec();
         }
 
-        public static bool DeleteLocal(string repo, string name)
+        public static bool DeleteLocal(string repo, string name, bool deleteNotMerged)
         {
             var cmd = new Command();
             cmd.WorkingDirectory = repo;
             cmd.Context = repo;
-            cmd.Args = $"branch -D {name}";
+            if (deleteNotMerged)
+            {
+                cmd.Args = $"branch -D {name}";
+            }
+            else
+            {
+                cmd.Args = $"branch -d {name}";
+            }
+
             return cmd.Exec();
         }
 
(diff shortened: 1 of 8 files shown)


love-linger:
* The confirmation panel to delete a branch itself is a reminder of this operation, the user needs to check and confirm, then click `OK` to start this operation.
* You can toggle `Enable --prune on fetch` from repository setting dialog.

ltetak:
I found the `Enable --prune on fetch,` thanks!

I don't fully agree that the confirmation panel is enough. How should I be sure that there are no pending changes? I would have to see if I pushed everything manually (one branch at a time). The -D option can cause changes to be lost, while the -d should be pretty safe and almost always do the job.

The problem I am trying to solve is cleaning up local folders that were merged. No git client has a great solution for this. Some at least notify you about the state of the branch (1. not-tracked 2. tracked 3. tracked but the origin branch has been deleted). This allows you to delete category 3 without risking deleting category 1.

love-linger:
I would like to add the track status to this panel if the target branch is local branch and has valid upstream and their HEAD is different. For example:

image

ltetak:
> I would like to add the track status to this panel if the target branch is local branch and has valid upstream and their HEAD is different. For example:
> 
That looks great. Then it should also go into the multi-delete dialog and the Local Branches list.
image

love-linger:
I'll close this issue since it is not necessary.

ltetak:
@love-linger I don't mind. I find it surprising that all git clients have `git branch -d` as default since it's the safer one. I think you should reconsider at least that.

love-linger:
I do not think so.

`Fork`, `GitKraken` and `JetBrains IDEs` use `-D` or `--delete --force` to delete a local branch. Even the `JetBrains IDE` doesn't require the user to confirm the operation again.

ltetak:
I am, obviously, not an expert so I won't argue. SourceTree (which resembles SourceGit the most) has -d as the default.
JetBrains Rider has a post notification with a rollback feature
image

PR #817, opened 2024-12-15
localization: add missing de_DE keys

NilsPvR:
Added missing keys and changed some instances of "Clone" with 'c' to a more German version with a 'k'. 

Took me a while to figure out what was meant with `InProgress.Merge.Operating` but I think the translation should be good now. 😅


love-linger:
> Added missing keys and changed some instances of "Clone" with 'c' to a more German version with a 'k'.
> 
> Took me a while to figure out what was meant with `InProgress.Merge.Operating` but I think the translation should be good now. 😅

`InProgress.Merge.Operating` used to label `Currently, we are processing the commit xxxxx`

PR #825, opened 2024-12-17
Change text selection brush to improve readability

mgarstenauer:
This PR changes the text selection brush for the AvaloniaEdit `TextEditor`.

### Problem
The text selection brush is the `SystemAccentColor`, which is a solid blue on Windows 11 by default. This can reduce the readability of the text when syntax highlighting is enabled.

### Solution
I suggest to use a semi-transparent brush to improve the readability.

### Before
image
image

### After
image
image


mgarstenauer:
The current opacity for the selection brush is 50%. We could reduce the opacity even more.

PR #836, opened 2024-12-24
New feature: system tray icon

wl2776:
Implements https://github.com/sourcegit-scm/sourcegit/discussions/805

## To enable

1. Open SourceGit's data storage directory from main menu
   image
2. Close SourceGit
3. Open file `preference.json` in any text editor
4. Add new field `"SystemTrayIcon": true`
   image

## Details

When this feature is enabled, SourceGit does not exit after click on window closing button.
Instead it stays in background, hiding in the system tray, without occupation of a space in the task bar.

image

Main window can be restored with the left click on the system tray icon.

Right click on the icon opens the menu with the Quit item.

## To disable

1. Open SourceGit's data storage directory from main menu
2. Close SourceGit with the Quit menu from the system tray icon
3. Open file `preference.json` in any text editor
4. Remove field "SystemTrayIcon" completely or set its value to `false`

### DIFF
--- a/src/App.Commands.cs
+++ b/src/App.Commands.cs
@@ -37,7 +37,7 @@ namespace SourceGit
             }
         }
 
-        public static readonly Command OpenPreferenceCommand = new Command(_ => OpenDialog(new Views.Preference()));
+        public static readonly Command OpenPreferenceCommand = new Command(_ => { OpenDialog(new Views.Preference()); });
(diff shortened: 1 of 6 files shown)


love-linger:
I do NOT like this feature at all.

wl2776:
OK, no problem.

Thank you very much for your work and for opening sources! 
I will keep this feature in my fork.

PR #845, opened 2024-12-27
localization: update

AquariusStar:
fix incorrected translated


AquariusStar:
Is it possible to create two translation options?
Otherwise, I know Russian programmers who got lazy and stopped doing correct and accurate translation and began to use transcription with the loss of the semantic load of translating some words and will be incomprehensible to another part of programmers, especially beginners and inexperienced.
Unfortunately, we will not be able to come to an agreement with the one who redid my translation. I suspect that there will be too much controversy and insults. It will interfere with your work. Therefore, I think it's better to create a translation option. And the people themselves will decide which translation will be more understandable or familiar by choosing the appropriate option.

Issue #566, opened 2024-10-14
Gitflow not working on version 8.34 in Mac OS


filipeRmlh:
The issue might be due to the change on how PATH enviroment variable are obtained in Mac OS.

SO: MacOS Sonoma
SourceGit Version: 8.34
Hardware: Macbook Pro - 2020 (intel x64)

-- With 8.33 I have no problem using gitflow.

Message: 
"git: 'flow' is not a git command. See 'git --help'.
The most similar commands are
	reflog
	show"

love-linger:
did you try the latest ci build from github action？

love-linger:
https://github.com/sourcegit-scm/sourcegit/actions/runs/11337556020

love-linger:
#539

Issue #545, opened 2024-10-08
[Regression Issue] - MacOS - sourcegit can't find npm global modules


yitzchak-ben-ezra-ecoplant:
Hi @love-linger 

Since you delivered this:  #507 I have an issue

the environment I'm using is this

1. shell: `zsh`
2. upon changing directory - `nvm` sets the correct `npm` version to be used
3. with the correct npm version I installed `pnpm` globally (`npm i -g pnpm`)
4. we have precommit hook that triggers something with `pnpm`
5. In order for source git to work with the correct `npm` version I'm launching it from `zsh` on the repo folder (to have the correct `npm` version) with this command: `open -a /Applications/SourceGit.app`

it used to work
now when I commit I see this message:
```
nx/.husky/pre-commit: line 4: pnpm: command not found
husky - pre-commit script failed (code 127)
husky - command not found in PATH=node_modules/.bin:/Applications/Xcode.app/Contents/Developer/usr/libexec/git-core:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
```

my full PATH from zsh is:
```
/Users/USERNAME/.pyenv/plugins/pyenv-virtualenv/shims:/Users/USERNAME/.pyenv/shims:/Users/USERNAME/.pyenv/bin:/opt/homebrew/opt/libpq/bin:/Users/USERNAME/.nvm/versions/node/v20.15.0/bin:/usr/local/share/dotnet:/opt/homebrew/bin:/Users/USERNAME/.local/bin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Applications/GPAC.app/Contents/MacOS/:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/Library/Apple/usr/bin:/Applications/VMware Fusion.app/Contents/Public:/usr/local/share/dotnet:~/.dotnet/tools:/Library/Frameworks/Mono.framework/Versions/Current/Commands:/Applications/iTerm.app/Contents/Resources/utilities:/Users/USERNAME/Library/Application Support/JetBrains/Toolbox/scripts
```

[...]

Issue #539, opened 2024-10-03
error: `env: node: No such file or directory`


BeLi4L:
When I try to commit some changes, I have the following error: `env: node: No such file or directory`as you see on this screenshot:

![image](https://github.com/user-attachments/assets/13cd53cc-9467-41de-9b8d-10f4c3c3f2f1)

I guess it's due to a pre-commit in my repo that uses yarn/node. Usually `node` is on my PATH, because I run `nvm` setup scripts in both .bashrc and .zshrc.

Any idea of why Sourcegit can't find `node`? Does it use a special shell or path? 🤔

love-linger:
it's because your sourcegit process is not started via bash or zsh. Are you using Sourcegit on macOS?

BeLi4L:
Yes I'm on macOS

love-linger:
The simplest way to solve this problem is changing the hook script's interpreter to `/bin/zsh`. For example:

```shell
#!/bin/zsh

node xxxxxx
```

aikawayataro:
`.*rc` files are only sourced when the shell is called interactively so to make it work with zsh, you should do PATH magic at `.zshenv'. There is no direct equivalent to bash though.

BeLi4L:
Thanks for your quick answers! ❤️ 

My hook scripts actually use `#!/bin/sh` so I just added `[ -s "/opt/homebrew/opt/nvm/nvm.sh" ] && \. "/opt/homebrew/opt/nvm/nvm.sh"` to every script, and it works 🥳

BeLi4L:
I noticed that **VS Code** doesn't have this issue (whereas sourcegit & some other git GUIs have it) @love-linger

Any idea why? Are they doing anything differently? 🤔 

I'm asking because to fix this in my company, I'd have to update every git hook in every repo, which is a huge pain, and may cause issues on some people's environments 😅

gadfly3173:
I guess it's because vscode reads various shell configuration files by itself.

https://github.com/microsoft/vscode/blob/3ae1a2259222266236891f983a0a9229b491998d/src/vs/workbench/contrib/terminal/common/scripts/shellIntegration-bash.sh

BeLi4L:
Not sure I understand 🤔 

Doing that would only allow me to access the `nvm` command, which isn't what I want, since the `nvm.sh` script is only a config script, which defines a `nvm` command, and adds the correct node/npm versions to PATH, which is what I want.

[...]

Issue #507, opened 2024-09-24
sourcegit could not find git-lfs in my path


htw128:
When pulling, I encountered the following error:

``` log
This repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting the 'post-checkout' file in the hooks directory (set by 'core.hookspath'; usually '.git/hooks').
```

Additionally, when switching branches, the switch command fails with the following error:

``` log
git-lfs filter-process: git-lfs: command not found
致命错误：远端意外挂断了
```

I checked in the terminal using which git-lfs, and it pointed to /opt/homebrew/bin/git-lfs.

The soucegit preferences are shown as follows:

<img width="604" alt="image" src="https://github.com/user-attachments/assets/272121f6-1369-491d-aef0-f451bcf94e6b">

Switching branches in the terminal works as usual. How do I fix this issue?

love-linger:
It's very strange. I've added the `/opt/homebrew/bin` to the `PATH` enviroment variable. You can open `/Applications/SourceGit.app/Contents/Info.plist` to check if the following settings exists in it?

```xml
<dict>
    <key>PATH</key>
    <string>/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
</dict>
```

love-linger:
similar to #168

htw128:
> It's very strange. I've added the `/opt/homebrew/bin` to the `PATH` enviroment variable. You can open `/Applications/SourceGit.app/Contents/Info.plist` to check if the following settings exists in it?
> 
> ```
> <dict>
>     <key>PATH</key>
>     <string>/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
> </dict>
> ```

Yes, it's in the Info.plist file from line 16 to line 19.

htw128:
I think it's worth mentioning that I'm running macOS Sequoia 15.0 (24A335).

love-linger:
I'm also using macOS Sequoia. Currently, you can run `open -a SourceGit` from terminal to avoid this problem. I'll try to fix it later

love-linger:
I found the right way to solve this problem.

1. run `launchctl setenv PATH $PATH`
2. restart your mac

love-linger:
[...]

PR #852, opened 2024-12-30
Support several more variants of git installation on MacOS

wl2776:
On MacOS git can be installed in `/usr/local/` or with HomeBrew.

Add support for these paths.

### DIFF
--- a/src/Native/MacOS.cs
+++ b/src/Native/MacOS.cs
@@ -25,7 +25,12 @@ namespace SourceGit.Native
 
         public string FindGitExecutable()
         {
-            return File.Exists("/usr/bin/git") ? "/usr/bin/git" : string.Empty;
+            var gitPathVariants = new List<string>() {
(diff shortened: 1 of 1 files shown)


TomaSajt:
By the way, why are we not using the `PATH` env-var, like done for Linux?
See https://github.com/sourcegit-scm/sourcegit/issues/72

love-linger:
> By the way, why are we not using the `PATH` env-var, like done for Linux? See #72

On macOS, GUI app cannot access the full `PATH` environment.
#566
#546
#545
#539
#507

wl2776:
Full list of global paths is stored in file `/etc/paths` and files under `/etc/paths.d/`

https://stackoverflow.com/a/38670834/704329

love-linger:
> Full list of global paths is stored in file `/etc/paths` and files under `/etc/paths.d/`
> 
> https://stackoverflow.com/a/38670834/704329

For example: you can see that the `/opt/homebrew/bin` is not list in `/etc/paths`, and `/etc/paths.d` only contains dotnet envs

image

PR #857, opened 2025-01-01
style(locale): Translate French locale further

UchiTesting:
- "Checkout" translates to "Récupérer". Updating lines that match.
- "Push" translates to "Pousser". Updating lines that match.
- Staged translates to "Présent dans l'index" but it is too long to I go simply for "Indexé".
- Translating a few extra lines.
- Some of the terms mentioned above remain untranslated because it is not appropriate provided their respective context.


UchiTesting:
Happy new year 2025! 🎉🎊🥳

May all your projects succeed and dreams come true ! 🥂

PR #868, opened 2025-01-04
System tray icon

wl2776:
Implements https://github.com/sourcegit-scm/sourcegit/discussions/805

# General description

When this feature is enabled, first launched instance of SourceGit doesn't quit when window closing button is pressed. Instead it hides in system tray and stays running.

This feature can be enabled or disabled either from "Preferences" dialog (menu -> Preferences -> Appearance) or by manually editing `preference.json` file.

# Implementation details

## Window hiding

Sourcegit intercepts `OnClosing` event, cancels it and calls `Hide` instead ([link to code](https://github.com/wl2776/sourcegit/blob/feature/system-tray-icon/src/Views/Launcher.axaml.cs#L261))

## Preventing from creation of several system tray icons

When another instance of the SourceHGit is launched from the command line or from the system launcher, it will create another identical icon in the system tray, if no additional checking is performed.

To prevent this, every SourceGit instance tries to create a lock file in the system temporary directory ([link to code](https://github.com/wl2776/sourcegit/blob/feature/system-tray-icon/src/Native/Linux.cs#L103)). If this operation fails, it means that another SourceGit is running and the system tray icon is already created. In this case, no system tray icons are created, and `OnClosing` event causes exit.

On MacOS [no lock file is created](https://github.com/wl2776/sourcegit/blob/feature/system-tray-icon/src/Native/MacOS.cs#L89), thus relying on the [OS guarantees to launch only one instance of a packaged application](https://github.com/AvaloniaUI/Avalonia/discussions/17854#discussioncomment-11700510).

# Why?

Because opening application that is already running is still much faster than launch a new instance.

# Note

Tested on Windows and Linux only.
MacOS is NOT tested.

### DIFF
--- a/src/App.Commands.cs
+++ b/src/App.Commands.cs
@@ -36,7 +36,8 @@ namespace SourceGit
 #endif
             }
         }
-
+        
+        public static readonly Command Unminimize = new Command(_ => ShowWindow());
         public static readonly Command OpenPreferencesCommand = new Command(_ => OpenDialog(new Views.Preferences()));
         public static readonly Command OpenHotkeysCommand = new Command(_ => OpenDialog(new Views.Hotkeys()));
         public static readonly Command OpenAppDataDirCommand = new Command(_ => Native.OS.OpenInFileManager(Native.OS.DataDir));
--- a/src/App.axaml.cs
+++ b/src/App.axaml.cs
@@ -16,6 +16,8 @@ using Avalonia.Data.Core.Plugins;
 using Avalonia.Markup.Xaml;
 using Avalonia.Media;
 using Avalonia.Media.Fonts;
+using Avalonia.Media.Imaging;
+using Avalonia.Platform;
 using Avalonia.Platform.Storage;
 using Avalonia.Styling;
 using Avalonia.Threading;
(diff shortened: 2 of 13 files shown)


wl2776:
Probably, there could be some minor issues in preferences dialog on Windows.
I would be grateful if someone could fix them.

wl2776:
Could anyone check how it works on MacOS? @love-linger? 

My goal is to make it staying in dock and keep running when window closing button is clicked.

wl2776:
Please, check this PR on MacOS

love-linger:
I'm sorry. As I said in Discussions #805, there's no plan to add this feature into `SourceGit`

wl2776:
Ok. You don't want to invest your time in this feature. I did it. Why not to use?

wl2776:
I have installed macos emulator and tested it myself. Currently doesn't work as intended. Converting to draft.

love-linger:
I don't think you understand what I mean. It's not because of that I don't have time to implement this feature, nor the feature is difficult to implement. Simply because `SourceGit` doesn't need this feature.

UsboKirishima:
I believe this feature has already been implemented in many other clients, and as a user, I would recommend implementing it as well. It is very convenient to have a program running in the background without needing to keep 256 tabs open, especially since the program is only used for a very small portion of the time during a coding session.

wl2776:
> It is very convenient to have a program running in the background without needing to keep 256 tabs open

Could you check how this PR works? I have tested it in Windows and in Linux, and it works as I have intended.

MacOS still needs work, since I don't have experience in macOS development. On macOS, I wanted SourceGit to stay in dock (in the bottom of the screen) instead of the system tray (top right corner). Anyway, the window is destroyed, for some reason.

PR #885, opened 2025-01-09
Remove Buildkite, add missing dependencies, README edits

aikawayataro:
Remove Buildkite workflow and README guide
Add libicu dependency for deb package
Add xdg-utils dependency for deb and rpm packages
Update Linux notes, xdg-utils now is dependency so xdg-open note is valid only for AppImage, few spelling edits


gadfly3173:
According to `dotnet-runtime-deps-9.0`, `libicu` needs to list more versions (for example, my system only has `libicu74`, which is not listed in this PR)
```
libc6, libgcc1, libgssapi-krb5-2, libstdc++6, zlib1g, libssl1.0.0 | libssl1.0.2 | libssl1.1 | libssl3, libicu | libicu74 | libicu72 | libicu71 | libicu70 | libicu69 | libicu68 | libicu67 | libicu66 | libicu65 | libicu63 | libicu60 | libicu57 | libicu55 | libicu52
```
image

aikawayataro:
> According to dotnet-runtime-deps-9.0, libicu needs to list more versions (for example, my system only has libicu74, which is not listed in this PR)

I took it from https://learn.microsoft.com/en-us/dotnet/core/install/linux-debian
Good catch, I see you fixed it in your PR

PR #890, opened 2025-01-10
build: Update package dependencies to include broader range of ICU

gadfly3173:


love-linger:
😂, Is there any other way to solve this problem? Currently, it looks unacceptable...

gadfly3173:
https://github.com/dotnet/runtime/blob/main/src/installer/pkg/sfx/installers/dotnet-runtime-deps/dotnet-runtime-deps-debian.proj dotnet also add libicu versions manually, I think there is no better way to fix this.

love-linger:
I checked AvaloniaUI's documentation and found that they also define the ICU dependencies in this way...

https://docs.avaloniaui.net/zh-Hans/docs/deployment/debian-ubuntu

PR #898, opened 2025-01-13
Exit on Ctrl+Q

wl2776:
Add hotkey Ctrl-Q for exit command on Linux and Windows.

Currently only MacOS supports exit on Cmd-Q.

### DIFF
--- a/src/Views/Hotkeys.axaml
+++ b/src/Views/Hotkeys.axaml
@@ -45,7 +45,7 @@
                    FontSize="{Binding Source={x:Static vm:Preferences.Instance}, Path=DefaultFontSize, Converter={x:Static c:DoubleConverters.Increase}}"
                    Margin="0,0,0,8"/>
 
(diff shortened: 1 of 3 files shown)


love-linger:
`Cmd+Q` is a common hotkey on macOS. The current implementation is only designed to comply with the operating habits of macOS. 

On Windows, users can use `Alt+F4` to do this, which is also a common hotkey to quit current application.

I don't have Linux device with me now, but I seem to remember that X11 also has a shortcut to close the current window by default

wl2776:
I am on Linux (Kubuntu 24.04), the application closes on Alt-F4.

Many other applications also support quit on Ctrl-Q.

love-linger:
I really do not want to break the operating system's rule, and I think it is really not necessary to provide multiple hotkeys to do the same thing.

wl2776:
In you terminology it is the rule.

image image  image  image image

PR #914, opened 2025-01-19
Add changed lines counter on commit details

### LINKED ISSUE #647: [Feature] Display number of lines added & removed
First of all I'd like to express my gratitude on the awesome work you've done so far in this project. After researching for a while for a good open source graphical git client, I found sourcegit, and it's awesome. Something I miss from sourcegit is the display of the number of lines removed and added in each commit, and in each file.

It would be something like this:

image

BernatBC:
- Fixes #647 

Added and removed lines are displayed on the commit details.
image

However, sometimes 0 is displayed, and correctly shown on second click. I am new to dotnet avalonia, and have no clue why is that happening.

I also plan to add these stats to files. However, I'm not sure where to place them (2 and 3)?:

image


gadfly3173:
> However, sometimes 0 is displayed, and correctly shown on second click. I am new to dotnet avalonia, and have no clue why is that happening.

Avalonia's TextBlock sometimes would not refresh correctly, see #338 and #383 .

PR #944, opened 2025-02-01
fix: adjust SHA regex pattern to min 6 chars

gadfly3173:
- Adjust the SHA regex pattern to match commit hashes with a minimum length of 6 characters instead of 10.

The minimum length of a Git commit ID can be 4 hexadecimal characters. Considering that `Diff.cs` uses a 6-40 character string for validation, the check for commit IDs in the commit message has been shortened to 6 characters. For example, GitHub considers a 7-character string as a valid commit ID.

### DIFF
--- a/src/Views/CommitMessagePresenter.cs
+++ b/src/Views/CommitMessagePresenter.cs
@@ -15,7 +15,7 @@ namespace SourceGit.Views
 {
     public partial class CommitMessagePresenter : SelectableTextBlock
     {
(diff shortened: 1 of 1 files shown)


love-linger:
It's because that some issue tracker system also uses 6-8 hex characters.

gadfly3173:
As far as I know, Gitee use `#I[5 characters]`, YouTrack and Jira use `[Project]-[number]`, and `CommitMessagePresenter` will not check the parts processed by `IssueTrackerRule` again. Or can we consider changing the check here to 7 characters?

Issue #940, opened 2025-01-30
view commit branch on graph


ghiboz:
during the scroll should be interesting have the information of the branch of the commit:

in alternative to show into the graph should be good have the info on the information panel

![Image](https://github.com/user-attachments/assets/a20d1b51-2773-417b-b800-c5ffa496e840)

gadfly3173:
> in alternative to show into the graph should be good have the info on the information panel

![Image](https://github.com/user-attachments/assets/bb9f489b-63ea-4801-86e7-72c0df5f4c2b) Actually the second button behind SHA can view branches

ghiboz:
> > in alternative to show into the graph should be good have the info on the information panel
> 
> ![Image](https://github.com/user-attachments/assets/bb9f489b-63ea-4801-86e7-72c0df5f4c2b) Actually the second button behind SHA can view branches

thank you! it's a not bad thing, but for example gitkraken shows directly on graph hovering the commits
that's more useful imho

gadfly3173:
![Image](https://github.com/user-attachments/assets/a498b83a-3df8-42aa-97a3-e3209873fecd)

#945 is my personal implementation, and the maintainers might have different ideas.

love-linger:
I don't want to make the commit graph so complicated. And the ability to query which REFs contains the selected commit is already available, I think that's enough.

PR #945, opened 2025-02-01
feat: show CommitRelationTracking on CommitGraph (#940)

gadfly3173:
- Add SHA property to the Dot class and assign commit SHA to anchor instances in the CommitGraph model.
- Add constructor to initialize commit relation tracking with repository path and commit hash, enabling querying and displaying related refs while managing loading state.
- Add pointer interaction handlers to the `CommitGraph` class for displaying commit details on click, updating cursor on hover, and resetting cursor on exit.

Only tested on Windows 10.

Image

### DIFF
--- a/src/Models/CommitGraph.cs
+++ b/src/Models/CommitGraph.cs
@@ -54,6 +54,7 @@ namespace SourceGit.Models
             public Point Center;
             public int Color;
             public bool IsMerged;
+            public string SHA;
         }
 
         public List<Path> Paths { get; } = [];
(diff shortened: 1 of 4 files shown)


love-linger:
I think the current implementation is good enough and there is no need to include this feature in the commit graph.

gadfly3173:
> I think the current implementation is good enough and there is no need to include this feature in the commit graph.

This idea was proposed by #940 and has two thumbs up. It might be better to comment in the issue.

Issue #786, opened 2024-12-04
[Enhancement] In File history add the capability of compare different versions of the same file


albertodlc:
I don't know if this is possible in other part of the App. I have a specific file version (in the File History menu) and I need to check the differences but using as reference different commits of the file.

![image](https://github.com/user-attachments/assets/84567504-2287-4e32-8e4d-cf05a49a7760)

love-linger:
currently, you can do this by selecting these two revisions(commits) in `History` view

love-linger:
currently, you can do this by selecting these two revisions(commits) in `History` view

albertodlc:
I see, I will use that for now. But if the commits are separated in the history menu or you just want to compare a single file it is easier to have the option in the `File history` directly.

The other way, first you need to select the ref. commit, then search or previously filter all the commits to locate the commit to compare. And then search in the full commit detail (between all the modified file) the specific file.

PR #946, opened 2025-02-02
feat: support compare both revision in FileHistories (#786)

gadfly3173:


src/Commands/CompareRevisions.cs
```diff
@@ -18,6 +18,15 @@ public CompareRevisions(string repo, string start, string end)
             Args = $"diff --name-status {based} {end}";
         }
 
+        public CompareRevisions(string repo, string start, string end, string path)
+        {
+            WorkingDirectory = repo;
+            Context = repo;
+
+            var based = string.IsNullOrEmpty(start) ? "-R" : start;
+            Args = $"diff --name-status {based} {end} -- {path}";
```
gadfly3173, line 27: I apologize, I just noticed that I forgot to add quotes around the `{path}`. Could you please fix this during your review?

PR #952, opened 2025-02-05
fix: Ignore the Chain-of-Thought in AI response

gadfly3173:
I reviewed the code of the official OpenAI SDK, and it seems that the official SDK does not handle the direct return of the Chain-of-Thought `<think>{thinking}</think>\n\n{content}`. After switching to the official SDK, you may need to perform additional processing on the returned results.

### DIFF
--- a/src/Commands/GenerateCommitMessage.cs
+++ b/src/Commands/GenerateCommitMessage.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 
 using Avalonia.Threading;
@@ -48,16 +49,18 @@ namespace SourceGit.Commands
                     var rs = new GetDiffContent(_repo, new Models.DiffOption(change, false)).ReadToEnd();
                     if (rs.IsSuccess)
                     {
+                        var hasFirstValidChar = false;
+                        var thinkingBuffer = new StringBuilder();
                         _service.Chat(
-                            _service.AnalyzeDiffPrompt, 
+                            _service.AnalyzeDiffPrompt,
                             $"Here is the `git diff` output: {rs.StdOut}",
                             _cancelToken,
                             update =>
-                            {
-                                responseBuilder.Append(update);
-                                summaryBuilder.Append(update);
-                                _onResponse?.Invoke("Waiting for pre-file analyzing to complated...\n\n" + responseBuilder.ToString());
-                            });
+                                ProcessChatResponse(update, ref hasFirstValidChar, thinkingBuffer,
+                                    (responseBuilder, text =>
+                                        _onResponse?.Invoke(
+                                            $"Waiting for pre-file analyzing to completed...\n\n{text}")),
+                                    (summaryBuilder, null)));
                     }
 
                     responseBuilder.Append("\n");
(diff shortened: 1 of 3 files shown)


gadfly3173:
move CoT removing method into GenerateCommitMessage to avoid remove think tags in prompts.

gadfly3173:
Improve regex pattern to specifically match and capture content within `<think>`, `<thought>`, `<thinking>`, or `<thought_chain>` tags.

gadfly3173:
Improve regex pattern to specifically match and capture content within `<think>`, `<thought>`, `<thinking>`, or `<thought_chain>` tags.

Can we merge this PR first? It removes some API service providers that do not support the openai `reasoning_content` field and adds CoT in the API response content.

Refer:
https://github.com/Mintplex-Labs/anything-llm/blob/master/frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/ThoughtContainer/index.jsx

https://github.com/CherryHQ/cherry-studio/blob/762c3d4950983f11f5e5552fdbfab95e99d3f929/src/renderer/src/utils/formats.ts#L90

love-linger:
The `openai-dotnet` team will release a new version within a week. Please wait for the `feat-openai` branch being merged first.

gadfly3173:
Can we merge this now? I want to use `deepseek-r1-distill-llama-70b` from groq, which is free and fast. 😄

love-linger:
I'll check this PR tomorrow

love-linger:
I'm using deepseek api (DeepSeek official service) with `deepseek-reasoner` model to test it, but I can not see the `<think>...</think` output...

gadfly3173:
> I'm using deepseek api (DeepSeek official service) with `deepseek-reasoner` model to test it, but I can not see the `<think>...</think` output...

The official API does not have this problem, but it may be encountered when deploying through open-source models, such as the `siliconflow` API that was just launched (it has now been modified) and `groq` (it seems they are not planning to change it). Discussions about this support can be found in the communities of various clients such as OpenWebUI, cherry-studio, etc.

gadfly3173:
> I'm using deepseek api (DeepSeek official service) with `deepseek-reasoner` model to test it, but I can not see the `<think>...</think` output...我正在使用 deepseek api（DeepSeek 官方服务）与 `deepseek-reasoner` 模型进行测试，但看不到 `<think>...</think` 输出...

You can try groq, which is free 1000 reqs/day

love-linger:
* In the next release's CHANGELOG, I will add the BREAKING CHANGE about the `Server` field of OpenAI configuration. I do not want to add some hard-code to migrate the old one. 
image

* How do you think that we just display the entire output of AI service, but remove the CoT while trying to apply it to the commit message?

gadfly3173:
> * In the next release's CHANGELOG, I will add the BREAKING CHANGE about the `Server` field of OpenAI configuration. I do not want to add some hard-code to migrate the old one.

I think remove `/chat/completions` is safe, also could avoid users input this by mistake. (It's highly unlikely that a real API provider would have an endpoint like "/chat/completions/chat/completions")

> * How do you think that we just display the entire output of AI service, but remove the CoT while trying to apply it to the commit message?

For most AI clients, it should be considered to separate the thinking chain from the display of answers. However, in the scenario where the AI returns results as submitted information here, the thinking chain is of no use. Moreover, the current processing results temporarily store the thinking chain in the `thinkingBuffer`, and it is also very easy to retrieve it later if someone really needs it.

love-linger:
All right. I have no questions. Good job!

gadfly3173:
Comments as records:

DeepSeek updated the chat template https://huggingface.co/deepseek-ai/DeepSeek-R1/commit/8a58a132790c9935686eb97f042afa8013451c9f , which makes the final output not include the starting `<think>` tag (which is really strange...) . For the current implementation, handling this situation can become cumbersome. `vLLM` has made additional processing and may need to observe how other API providers and chat clients will react.

PR #956, opened 2025-02-06
fix: when upstream is null show CommitAndPush and stop auto push (#955)

### LINKED ISSUE #955: [Feature Request] Add commit and push option when remote branch does not exists
If a remote branch exists, the option to make a commit and push it appears, but not when the remote branch was never created.

I want the option always to have the option of commit & push, and in that case, make the same behaviour as if you make a push to origin manually.

Without a remote branch:
Image

With a remote branch:
Image

gadfly3173:
Prevent the automatic execution of the push operation when the `NEW` label is visible. This would ensure that users have the opportunity to confirm the creation of a new branch on the remote, rename the branch, or cancel the operation.

Fixes #955

### DIFF
--- a/src/ViewModels/LauncherPage.cs
+++ b/src/ViewModels/LauncherPage.cs
@@ -59,14 +59,16 @@ namespace SourceGit.ViewModels
         public void StartPopup(Popup popup)
         {
             Popup = popup;
-            ProcessPopup();
+            ProcessPopup(true);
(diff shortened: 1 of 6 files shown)


gadfly3173:
new `AutoCheck` method is required. Because manually open `Push` popup and click `Sure` button, `Check` method also will be called.

PR #967, opened 2025-02-10
Fix: Swap selection start and end index if selection is created from bottom to top

### LINKED ISSUE #966: [Regression] Can only copy from file diff when selecting with Ctrl-a
I can only copy from file diff view when using Ctrl-a to select everything. Both Ctrl-c and right click > copy work then.

I cannot copy when selecting lines with mouse/keyboard, neither with Ctrl-c nor with right click > copy.

I think this might be a regression related to #924.

Doodeletion:
Fixes #966.

Copying from file diff view failed if the selection is created from bottom to top. This causes Selection.StartPosition to be below Selection.EndPosition, so the for loop would not run. Swapping start and end position in these cases fixes the issue.

On a different note, I noticed that any exception thrown in this event handler crashes the application. I think in some cases, such as this, it might be better to catch and log the exception and prompt the user about it.

### DIFF
--- a/src/Views/TextDiffView.axaml.cs
+++ b/src/Views/TextDiffView.axaml.cs
@@ -970,8 +970,19 @@ namespace SourceGit.Views
             }
 
             var lines = GetLines();
-            var startIdx = Math.Min(selection.StartPosition.Line - 1, lines.Count - 1);
-            var endIdx = Math.Min(selection.EndPosition.Line - 1, lines.Count - 1);
+
+            var startPosition = selection.StartPosition;
+            var endPosition = selection.EndPosition;
+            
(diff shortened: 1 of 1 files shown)


gadfly3173:
> On a different note, I noticed that any exception thrown in this event handler crashes the application. I think in some cases, such as this, it might be better to catch and log the exception and prompt the user about it.

Actually, any exception in the UI thread or any uncaught exception in other threads can cause Avalonia to crash... The non-UI thread part has been handled, but for the UI thread, there is a considerable part that Avalonia itself has not handled (such as the rendering of AvaloniaEdit). For this specific case of copying a selection, it should not throw an exception.

PR #981, opened 2025-02-14
feat: Push to any remote branch

Muszkal:
When working on a branch, You either push a branch to the remote(which might exist or not) with the same name or a new branch completely.
I don't think there are situations where you push to an existing remote branch that is different from the one you are on currently.

This solution gives the flexibility to push to any branch you want.

My usecase:
I develop on "develop" branch, than push to "feature" branch.
This is useful when dealing with bugs or small things.

before:
image

after:
image

and with new branch:

before: 
image

after:
image

### DIFF
--- a/src/ViewModels/Push.cs
+++ b/src/ViewModels/Push.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
+using System.Linq;
 using System.Threading.Tasks;
 
 namespace SourceGit.ViewModels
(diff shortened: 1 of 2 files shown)


love-linger:
Instead, I recommend creating the corresponding branch locally first, and then pushing it directly. For example: use a workflow tool such as `git-flow`. Always ensuring that the local branch name matches the remote branch name it tracks (the upstream of local branch) helps to understand the entire repository.

Imagine a situation where you have a `develop` branch both locally and remotely. If you add new changes directly to it, and then push it directly to another new remote branch `feature/xxxx` and then wait for the maintainer to merge into the `develop` branch. At this time, there are new requirements to be developed on the `develop` branch. Do you need to reset the `develop` branch to the original revision (considering that the previous `feature/xxxx` may not be merged into the develop branch by the maintainer for some reason)?

Muszkal:
If after I pushed `develop` into `feature/xxx`, then need to develop something on the `develop` branch, then I would reset to remote.
When starting to develop something new or a bug fix, always start from resetting to remote.

If after pushing `feature/xxx`, I need to change something then I would checkout `feature/xxx`

Muszkal:
There are more use cases for the ability to push to any branch I chose.
For example if i have a `feature/xxx` branch and I decide to try something on it, than I would create a `feature/newxxx` branch to test and if the test is successful than I would push `feature/newxx` to `feature/xxx`

love-linger:
> There are more use cases for the ability to push to any branch I chose. For example if i have a `feature/xxx` branch and I decide to try something on it, than I would create a `feature/newxxx` branch to test and if the test is successful than I would push `feature/newxx` to `feature/xxx`

Why not merge/reset/rebase the `feature/newxx` branch locally first and then push the merged/reset/rebase `feature/xxx` to the remote?

Muszkal:
Well, I am not saying it is not possible. I would like flexibility so that different people with different workflows have a good experience. This change gives that flexibility without taking out something else.

On a side note, Is there a discord or some form of chat for this project?

mithom:
i'm in favor with the love-linger here.
GUI's are there to make working with git easier.
Supporting bad habbits does not make git easier for beginners.
if you want all flexibility git offers, there is the CLI, or sourceTree for you. i believe that GUI makers have the duty to guide users to use their tool correct.
also an easy solution that would keep the good habbits in place is to just rename your local branch and then push, keeping the names in sync.

PR #1006, opened 2025-02-19
management of local branch with deleted remote

dtentori:
local branches with deleted remote (no upstream) are marked with "warning" icon

### DIFF
--- a/src/Commands/QueryBranches.cs
+++ b/src/Commands/QueryBranches.cs
@@ -14,7 +14,7 @@ namespace SourceGit.Commands
         {
             WorkingDirectory = repo;
             Context = repo;
-            Args = "branch -l --all -v --format=\"%(refname)%00%(objectname)%00%(HEAD)%00%(upstream)%00%(upstream:trackshort)\"";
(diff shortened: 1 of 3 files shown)


love-linger:
I've pushed another implement of this feature. See 53f591bdad68afeb49a4c6d0b0ea8fc42b7a7323

TomW-Skyline:
@love-linger is there an easy way to (automatically) remove all branches where the remote branch is gone?

love-linger:
@TomW-Skyline There's no plan to add such a feature. The purpose of using `-d` instead of `-D` when deleting branch is that this is a dangerous operation. You can only manually selecting all the branches you want to delete (use Ctrl to select multi-branches)

TomW-Skyline:
> @TomW-Skyline There's no plan to add such a feature. The purpose of using `-d` instead of `-D` when deleting branch is that this is a dangerous operation. You can only manually selecting all the branches you want to delete (use Ctrl to select multi-branches)

@love-linger When using ctrl to select multiple branches, I'm getting the "is not fully merged" error without possibility to force delete them.

love-linger:
@TomW-Skyline I've pushed a new commit that adds an option to enable `--force` when deleting multiple branches

Issue #995, opened 2025-02-17
[Feature] Add a prune checkbox to the fetch dialog and remember its state


Autre31415:
Currently the prune option can only be set globally in the settings screen.

love-linger:
...

1. Before `8.35`, there's a `Prune remote dead branches` in `Fetch` popup
2. In `8.35`, the `Prune remote dead branches` has been removed. #590
3. In `8.37`, I added a pre-repository configuration `Enable --prune on fetch`. #590
4. In `2025.04`, the `Enable --prune on fetch` has been moved to global git configure `fetch.prune`. #980

love-linger:
I've pushed a new commit that supports configure `fetch.prune` for a single repository from `Repository Configura` window.

You can download the latest CI build from Github Action.

PR #1010, opened 2025-02-20
feat: Added option to prune outdated refs during fetch

kosmakoff:
This change adds options to Fetch dialog to prune outdated remote references, i.e. those remotes that do not exist anymore.

### DIFF
--- a/src/Commands/Fetch.cs
+++ b/src/Commands/Fetch.cs
@@ -4,7 +4,7 @@ namespace SourceGit.Commands
 {
     public class Fetch : Command
     {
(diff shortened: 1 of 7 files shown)


love-linger:
#995 

You can enable global `fetch.prune` in `Preference` window. 

For these reasons, this PR will not be merged

* This PR is the same as the implementation of `SourceGit` before `8.35`. The reason why it is changed to the current state is due to some feedback and consideration.
* In this PR, when `Prune outdated remote refs` is unchecked, if `fetch.prune` is enabled in global/local(repository), the actual behavior does not match the design

PR #1049, opened 2025-03-04
fix: Prevent extra newlines when copy lines

gadfly3173:


love-linger:
I've pushed another fix for this issue.

PR #1059, opened 2025-03-06
Enable tooltips for the commit hashes when handling conflicts

DarkDefender:
Before, you couldn't easily see the commit hash authors or message when resolving file conflicts. Now you can hover over the hashes to get a tooltip with a bit more information.

I have to admit that this is my first time working with C# and Avalonia, so I'm not sure how hacky my solution is... :sweat_smile:

### DIFF
--- a/src/ViewModels/Repository.cs
+++ b/src/ViewModels/Repository.cs
@@ -821,6 +821,11 @@ namespace SourceGit.ViewModels
             }
         }
 
+        public Models.Commit GetCommitInfo(string sha)
+        {
+            return new Commands.QuerySingleCommit(_fullpath, sha).Result();
+        }
+
         public void NavigateToCurrentHead()
         {
             if (_currentBranch != null)
--- a/src/Views/WorkingCopy.axaml
+++ b/src/Views/WorkingCopy.axaml
@@ -216,7 +216,23 @@
                                      Foreground="DarkOrange"
                                      TextDecorations="Underline"
                                      Cursor="Hand"
-                                     PointerPressed="OnPressedSHA"/>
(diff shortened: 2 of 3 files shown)


love-linger:
* First, we already shows the tooltip in `In-Progress Bar`
image

* It's better to supports to navigate to selected commit just like another well-known GIT client `Fork`

DarkDefender:
> It's better to supports to navigate to selected commit just like another well-known GIT client Fork

I agree! I see you already added it as well! Thanks!

> First, we already shows the tooltip in In-Progress Bar

Ah, I missed that!
Wouldn't it be good to show tooltips when hovering over the commit hashes in the conflict area as well though?

Feels weird to me that you can click on them now but not hover over them (like you can for the parent hash in the commit information view).

DarkDefender:
One of the reasons I added the hover tooltop is because the `In-Progress Bar` doesn't show up for stash conflicts.
I have a change for making for info show up about stashed changes here: https://github.com/sourcegit-scm/sourcegit/issues/1067

Even if there was a `In-Progress Bar` for stash conflicts, I would still like to also be able to hover over the hashes in the conflict view. I just wanted to add a bit more context to why I created this PR.

DarkDefender:
I've cleaned up my changed to be much more simpler like in your changes! Thanks for showing me how it is done. :)

Now the hashes work just like in the "commit message" window.

love-linger:
I've pushed my implementation for this feature. Because of some commits already exist in `SourceGit.ViewModels.Conflict`, it not necessary to fetch them on first-time-hover.

DarkDefender:
Thank you so much!

PR #1060, opened 2025-03-06
fix: Show detached HEAD

xhargh:
I noticed that a detached HEAD wasn't shown properly. Here is my try to fix it.

### DIFF
--- a/src/ViewModels/Repository.cs
+++ b/src/ViewModels/Repository.cs
@@ -1012,7 +1012,7 @@ namespace SourceGit.ViewModels
 
             var filters = _settings.BuildHistoriesFilter();
             if (string.IsNullOrEmpty(filters))
-                builder.Append("--branches --remotes --tags");
+                builder.Append("--branches --remotes --tags HEAD");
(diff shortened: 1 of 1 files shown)


love-linger:
* First, I do not understand why `HEAD` and tags are always filtered in `git log` command
* Second, I have not notice the difference behaviour between this PR and `develop` branch

love-linger:
> I noticed that a detached HEAD wasn't shown properly

image

xhargh:
I was not clear enough, my change fixes when we have a HEAD detached from something, like this:

image

which was shown like this before:

image

love-linger:
I see. Then you should only add it to the default options if no filter is specified, such as:
image

love-linger:
By the way, without this PR, you can find it by enable `--reflogs`
image

xhargh:
I think your version with adding HEAD if there are no filters is better.

I know about the reflog option, that is how I've even found the HEAD earlier. But in the reflog, there are other things that are not always of interest

PR #1069, opened 2025-03-07
Working copy status refresh indication and index lock preventive handling

bogomolets-owl:
This merge request adds two new minor features:
- Showing local changes update status with a separate activity indicator.
- Running commands aware of `index.lock` behavior.

Both features aim to enhance the UX when working with large git repositories, particularly those containing a lot of files. Large file counts make `git status` a lengthy operation. Without these changes, SourceGit attempts to execute commands that fail on the git side because of the existence of `index.lock`. Also user dont see when the status update is finished and the local changes state is up to date.

### DIFF
--- a/TRANSLATION.md
+++ b/TRANSLATION.md
@@ -6,7 +6,7 @@ This document shows the translation status of each locale file in the repository
 
 ### ![en_US](https://img.shields.io/badge/en__US-%E2%88%9A-brightgreen)
 
-### ![de__DE](https://img.shields.io/badge/de__DE-98.79%25-yellow)
+### ![de__DE](https://img.shields.io/badge/de__DE-98.65%25-yellow)
 
 <details>
 <summary>Missing keys in de_DE.axaml</summary>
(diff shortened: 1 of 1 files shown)


love-linger:
* First, for read-only git operations, such as `git status`, `index.lock` will not fail that command.
* I do NOT think it is better that handle the `index.lock` manually, since the `index.lock` file itself is a protection measure for repository data in git.

bogomolets-owl:
The scenario i want to fix is:
- user focuses on SourceGit window
- SourceGit starts git status operation, that locks the repo. This operation can take 10 to 45 seconds (depends on untracked files setting) on the repository.
- user starts git pull operation. 
- git pull operation fails due to git status index.lock 
In this case user have no clear indication on git status progress and no clear understanding that repository is currently locked.

love-linger:
I've pushed a new commit 2b2f070c4aae5789f848a8fe56dead45cd425c34 that enable `--no-optional-locks` for `git status` command.

bogomolets-owl:
This indeed seems as a cleaner solution for the status locking issue, thank you!

Though the lengthy git status times remain an issue.
To understand the scenario:
- User focuses SourceGit window
- SourceGit starts git status operation that takes 10-45 secs.
- User examine working copy changes. 
In this case some changes will be missing until the status command is finished and user have no clear indication about working copy status update is still in progress. 

I will update the code and remove any Commands code modifications so only the visual indication remains.

PR #1078, opened 2025-03-10
fix: Force English locale in branch query command.

gadfly3173:
New:
image

Old:
image


gadfly3173:
`en_US.UTF-8` not work on my Deepin , may because of I didn't install English locale in system. `LANG=C` has better compatibility

PR #1092, opened 2025-03-13
fix: improve font string processing in SetFonts method

gadfly3173:
Fixed the crash when the font is `,SomeFont`.

Situation encountered:
When I added `Noto Color Emoji`, I wrote it at the beginning of the default font. In the TextBox, it was `Noto Color Emoji, HarmonyOS Sans SC`. Then I found that the digital font became emoji, so I deleted the content before `,` and the whole app crashed.

### DIFF
--- a/src/App.axaml.cs
+++ b/src/App.axaml.cs
@@ -181,6 +181,9 @@ namespace SourceGit
                 app._fontsOverrides = null;
             }
 
+            defaultFont = ProcessFontString(defaultFont);
+            monospaceFont = ProcessFontString(monospaceFont);
+
             var resDic = new ResourceDictionary();
             if (!string.IsNullOrEmpty(defaultFont))
                 resDic.Add("Fonts.Default", new FontFamily(defaultFont));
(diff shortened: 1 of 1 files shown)


love-linger:
I suggest you to move your code into `SourceGit.ViewModels.Preferences.FixFontFamilyName`

https://github.com/sourcegit-scm/sourcegit/blob/e3cc9876824f54ed24f0833d87a6b3b260c3784c/src/ViewModels/Preferences.cs#L617-L644

gadfly3173:
If `move code into SourceGit.ViewModels.Preferences.FixFontFamilyName` , It would be unable to type `,` at the end of the font string

love-linger:
If we do not remove the whitespace char at the end of a font name, `SourceGit` will also crashed.

Try `Microsoft YaHei , Consolas`

love-linger:
I'll merge this PR first, and add some modifications.

gadfly3173:
https://github.com/sourcegit-scm/sourcegit/blob/03681aef8c091930670c38849811fbee10519ec1/src/App.axaml.cs#L453

my code trimed every font

PR #1096, opened 2025-03-14
feat: add translation for "1 hour ago" in multiple languages

ousugo:
PixPin_2025-03-14_16-47-49


gadfly3173:
Maybe `an hour ago` would be better in English?

https://github.com/moment/moment/blob/18aba135ab927ffe7f868ee09276979bed6993a6/src/lib/locale/relative.js#L1-L18

love-linger:
I think it does not really matter. See https://forum.wordreference.com/threads/one-hour-or-an-hour.2693476/

PR #1106, opened 2025-03-17
fix: improve line splitting to handle both LF and CRLF line endings

gadfly3173:


src/Commands/QueryCommitSignInfo.cs
```diff
@@ -22,7 +24,7 @@ public Models.CommitSignInfo Result()
             if (raw.Length <= 1)
                 return null;
 
-            var lines = raw.Split('\n');
+            var lines = raw.Split(['\n', '\r'], StringSplitOptions.RemoveEmptyEntries);
```
love-linger, line 27: You can NOT use `StringSplitOptions.RemoveEmptyEntries` option here, because `Signer` may be missing.

love-linger:
I'll try to fix this

gadfly3173:
src/Commands/QuerySubmodules.cs `\r` 被改成了 `r`

Issue #1121, opened 2025-03-24
[Enhancement] Add loading animation for switching workspaces


gadfly3173:
When switching workspaces, the UI will get stuck. Even if you minimize the window, the context menu will remain on the screen until the workspace switch is complete. It seems that opening the repository is a time-consuming operation. I hope to add a loading animation to the function of switching workspaces.

![Image](https://github.com/user-attachments/assets/33b68cb5-a692-43bc-b573-b059f201368e)

```log
[SwitchWorkspace] Starting workspace switch to: xxxx
[SwitchWorkspace] Task check completed: 3ms
[SwitchWorkspace] Workspace states updated: 6ms
[SwitchWorkspace] Existing pages cleared: 111ms
[SwitchWorkspace] 5 repositories reopened: 1164ms
[SwitchWorkspace] Active page set: 1169ms
[SwitchWorkspace] Preferences saved: 1177ms
[SwitchWorkspace] Completed in 1205ms
```

love-linger:
It's because we need get the `$GIT_DIR` before opening a repository.

![Image](https://github.com/user-attachments/assets/cde80867-9383-45c6-98c5-f50303c11254)

These two commands took up most of the time. You can test it by the following code

```diff
- gitdir = new Commands.QueryGitDir(node.Id).Result();
+ gitdir = Path.Combine(node.Id, ".git");
```

gadfly3173:
> I'm not sure if the follow code is a right way to solve this problem:

```log
[SwitchWorkspace] Task check completed: 6ms
[SwitchWorkspace] Workspace states updated: 10ms
[SwitchWorkspace] Existing pages cleared: 103ms
[SwitchWorkspace] 5 repositories reopened: 1329ms
[SwitchWorkspace] Active page set: 1333ms
[SwitchWorkspace] Preferences saved: 1336ms
[SwitchWorkspace] Completed in 1379ms
```

Doesn't seem to be improving.

BTW, I noticed that switching workspaces when the repository is not fully loaded may cause some AggregateExceptions, probably because the repository is closed during the execution of some `Task.Run`.

```log
System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (Object reference not set to an instance of an object.)
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
[...]

PR #1139, opened 2025-03-31
refactor: improve auto fetch timer implementation using PeriodicTimer

gadfly3173:
#792 #1121

Replace System.Threading.Timer with PeriodicTimer for auto fetch functionality to:
- Handle cancellation gracefully during repository close
- Avoid unhandled exceptions in timer callbacks


love-linger:
why not just using `try .. catch` statement in `AutoFetchImpl`?

gadfly3173:
Reasonable. I'm thinking too much.

PR #1153, opened 2025-04-04
Add filterbox in Staged area, same as in Unstaged area

goran-w:
Implements #1155.

Adds a searchfilter-box, identical to the one in the Unstaged area.

Also adds a RaiseException if committing with a Staged-filter active, since otherwise more changes than the visible ones could be committed. NOTE: This could be made into a warning-dialog with options "Continue? Yes / No", similar to `ConfirmCommitWithoutFiles()`.


love-linger:
I'll check this PR later.

PR #1158, opened 2025-04-06
Locale fr 2504

UchiTesting:
Took a little time to continue with French translation.

Is this correct in English version ?

`<x:String x:Key="Text.SetUpstream" xml:space="preserve">Set Tracking Branch</x:String>`

Key tells upstream but value tells tracking branch. I followed the value.


love-linger:
You are right. At first, this function was `Set Upstream` since the backend git command is `git branch --set-upstream`, but after referring to some other Git clients, I found that they usually use `tracking branch` to represent this relationship, so I changed it to `Set Tracking Branch`

Issue #1161, opened 2025-04-07
[BUG] Duplicate entry on french translation.


glaplace:
Hello,

from the last version sourcegit does not start with error

System.ArgumentException: An item with the same key has already been added. Key: Text.CommitCM.Checkout
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey, TValue, InsertionBehavior) + 0x297
   at CompiledAvaloniaXaml.!AvaloniaResources.Populate:/Resources/Locales/fr_FR.axaml(IServiceProvider, ResourceDictionary) + 0x12c3
   at SourceGit.App.XamlClosure_1.Build_3(IServiceProvider) + 0x6a
   at Avalonia.Markup.Xaml.XamlIl.Runtime.XamlIlRuntimeHelpers.DeferredContent`1.Build(IServiceProvider serviceProvider) + 0xd7
   at Avalonia.Controls.ResourceDictionary.TryGetValue(Object, Object&) + 0x87
   at Avalonia.Controls.ResourceDictionary.get_Item(Object) + 0x10
   at SourceGit.App.SetLocale(String localeKey) + 0x78
   at SourceGit.App.Initialize() + 0x6f
   at Avalonia.AppBuilder.SetupUnsafe() + 0x10d
   at Avalonia.ClassicDesktopStyleApplicationLifetimeExtensions.StartWithClassicDesktopLifetime(AppBuilder, String[], Action`1) + 0x34
   at SourceGit.App.Main(String[] args) + 0xe0

the key Text.CommitCM.Checkout are duplicated since the commit 904432a8f1d63472f913071ad1e52f5a8e5fa582 

another key are duplicated 
Text.StashCM.SaveAsPatch line 619 and 765
Text.Stash.AutoRestore.Tip ligne 763 and 609
Text.Stash.AutoRestore 609 and 762
Text.SHALinkCM.NavigateTo  line 602 and 761
Text.SetUpstream.Upstream 600 and 760
Text.SetUpstream.Unset 599 and 759
Text.SetUpstream.Local line 598 and 760
Text.SetUpstream line 597 and 757
Text.CreateBranch.Name.WarnSpace ligne 200 and 714

Text.CommitCM.Checkout line 103 and 101
Text.BranchUpstreamInvalid line 64 and 710
Text.BranchCM.CustomAction line 48 and 709
Text.ApplyStash.Stash line 708 and 31
Text.ApplyStash.RestoreIndex line 707 and 30
[...]

PR #1163, opened 2025-04-07
fix(#1161): Fix French translation

glaplace:


love-linger:
I've pushed a commit that fixed this issue. See f5c213060eb842a302713ba27178c0f0a0c6e37b

PR #1177, opened 2025-04-11
refactor: improve diff handling for EOL changes and enhance text diff…

massimopaganigh:
… display

- Updated `Diff.cs` to streamline whitespace handling in diff arguments.
- Enhanced `DiffContext.cs` to check for EOL changes when old and new hashes differ, creating a text diff if necessary.
- Added support for showing end-of-line symbols in `TextDiffView.axaml.cs` options.

### DIFF
--- a/src/Commands/Diff.cs
+++ b/src/Commands/Diff.cs
@@ -28,9 +28,9 @@ namespace SourceGit.Commands
             Context = repo;
 
             if (ignoreWhitespace)
-                Args = $"-c core.autocrlf=false diff --no-ext-diff --patch --ignore-cr-at-eol --ignore-all-space --unified={unified} {opt}";
+                Args = $"-c core.autocrlf=false diff --no-ext-diff --patch --ignore-all-space --unified={unified} {opt}";
             else
-                Args = $"-c core.autocrlf=false diff --no-ext-diff --patch --ignore-cr-at-eol --unified={unified} {opt}";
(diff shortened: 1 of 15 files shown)


massimopaganigh:
I thought about implementing these changes because I often commit without realising that I have inadvertently changed all the EOFs. I was inspired by what Git Extensions does, which is to report EOFs as changes if you choose to show whitespace.

P.S.
I had some help with the translations, but still checked that the changes made sense, so I apologise to the translators in advance if I made any mistakes.

love-linger:
Please remove the changes about `Text.Apply.IgnoreWS` translations. It's not used by diff viewer.

massimopaganigh:
> Please remove the changes about `Text.Apply.IgnoreWS` translations. It's not used by diff viewer.

Alright

love-linger:
Since some people have reported that they are not used to the new `EOL` processing way, and some commonly used tools such as `Github` and `Fork` also enable `--ignore-cr-at-eol` by default, this PR will be reverted in the next version.

love-linger:
Another reason is that `Options.ShowEndOfLine = true` shows the line-endings in the **diff command output**, and may not be the line-endings of the original file. 

For example, if there is no empty line at the end of the file before the change, the diff result will have an extra line `\ No newline at end of file`. This causes the last line of the original file to have an extra line-ending.

PR #1179, opened 2025-04-13
using the new syntax

qiufengshe:
using the new syntax


love-linger:
I'm sorry. There's no plan to use the new `new` syntax

PR #1209, opened 2025-04-17
feature: add "Copy Subject" option to commit context menu

ousugo:
PixPin_2025-04-17_18-29-47


love-linger:
I suggest you to use `Copy Info` instead.

ousugo:
My usage scenario is that I often select the title of a certain commit as the title of a PR. In this case, the SHA value copied at the beginning using `copy info` is rather inconvenient

love-linger:
I've pushed my implementation of this feature request.

image

ousugo:
look great, thanks

Issue #1210, opened 2025-04-17
[Feature Request] Popup for merge conflicts


lwray-renesas:
Hello,

First of all, thanks for such a great tool!
I have a minor suggestion that might be good for first time users.

FYI I'm using Windows 10.
When in the HISTORY view and trying to merge changes between branches which contain conflicts that can't be resolved automatically, the details show that a the merge is in progress, indicating the merge is incomplete.
However for new users, its not immediately obvious that there are unresolved conflicts and that they need to go to LOCAL CHANGES and handle conflicts.

Suggestion is:
On detection of unresolved conflicts, to produce a pop-up/dialog that indicates conflicts exist which haven't been resolved and offers the user a button to launch their external merge tool on the files affected.

Thanks for any consideration!

Louis

netodotcom:
This suggestion is good. I'm experiencing the same thing.

love-linger:
I suggest you to use a external merger.

![Image](https://github.com/user-attachments/assets/fc7be591-4288-42dd-88a1-40b36521040b)

love-linger:
You can also open external merger for all conflicts one-by-one by clicking this button.

![Image](https://github.com/user-attachments/assets/eb4fba23-ab6f-4ea8-ab78-c2fbc736042e)

lwray-renesas:
Thanks for the quick response - Sorry, I don't think my request was clear.
Its not that I don't know how to launch the external merge tool, its that after I perform a merge and conflicts exist (whilst in the HISTORY view) its not immediately obvious that there are conflicts pending for a new user.

![Image](https://github.com/user-attachments/assets/3b721492-48bc-449e-8455-e836f60ca47d)

I have to already know that "Merge in progress" means I need to navigate to LOCAL CHANGES and start addressing the conflicts there.

My request is that when a conflict is detected, a dialog popping up to indicate there is a conflict would help new users catch it.
Another suggestion might be to default the view to LOCAL CHANGES and select the first visible unstaged change, forcing the interface to the conflicts detected messaging.

Code wise, something like:

![Image](https://github.com/user-attachments/assets/da4d68c0-f2a1-4d2f-a4d0-1882ae092a3c)

lwray-renesas:
Just managed to mock up the behaviour if it helps illustrate?
Effectively forcing the user to deal conflicts, codes here if you want to give it a go, I'll open a pull request shortly and link here:
```c#
Dispatcher.UIThread.Invoke(() =>
            {
[...]

PR #1211, opened 2025-04-17
Feat/conflict popup

lwray-renesas:
Addressed feature request https://github.com/sourcegit-scm/sourcegit/issues/1210

Now when there is a conflict (either on startup or after an attempted merge) that cannot be resolved, a popup appears and the view switches to the first conflict to resolve.

image

image

Also, previously tool tip erroneously repeated --theirs when hovering over USE MINE
Now it correctly says --ours:

image


love-linger:
I've picked the commit https://github.com/sourcegit-scm/sourcegit/pull/1211/commits/26a471933c21a135e946273d49b9135b401bbd7f manually.

I do not like the `Conflict-Popup` design. See #1210

And I do not like the idea that always auto-selecting the `LOCAL CHANGES` page when there're conflicts. There're two reasons:

* In-progress operation may not only due to conflicts, but also empty commits or waiting user to edit commit in `interactive-rebase`.
* In this PR, every time after `SetData` being called (generally after `SourceGit` detected the files in your worktree have been changed), the `LOCAL CHANGES` will be selected if there are conflicts. Imagine the following situation: when a conflict occurs, the user switches to the `HISTORY` or `STASHES` page and copies some content from a certain revision of the file to resolve the conflict. But every time after copying, SourceGit will detect file changes and forcefully return to the `LOCAL CHANGES` page automatically.

PR #1224, opened 2025-04-21
minimize temporary strings for better performance

qiufengshe:
reduce temporary strings, reduce memory allocation, and improve performance


gadfly3173:
@love-linger this PR has been merged into `master` instead of `develop`.

PR #1243, opened 2025-04-25
Use ComboBox to select font

fallenwood:
It uses text input for fonts now

Image

It will be easier to use if use ComboBox to select system fonts

Image

Tested on windows 11

### DIFF
--- a/src/ViewModels/Preferences.cs
+++ b/src/ViewModels/Preferences.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 using Avalonia.Collections;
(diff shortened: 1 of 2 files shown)


love-linger:
This is deliberately designed this way. And in previous versions, SourceGit used a `ComboBox` to display font effects (the FontFamily of each item was bound to its corresponding font). The reasons for changing to the current implementation method are as follows:

1. For some machines with relatively low configurations, it takes a long time to load the font list, and the interface will be unresponsive for a while. For example, some Linux systems come with a large number of `NotoSans` series fonts installed by default.
2. Excessive memory consumption.
3. The current implementation method supports combined fonts. For example, if you fill in: `Segoe UI, Microsoft YaHei UI`, it will give priority to using the first font. When a missing character is encountered, it will search the others.

PR #1265, opened 2025-05-01
feat: Add "Open in Browser" button to toolbar

L1cardo:
- Add "Open in Browser" button to toolbar.
- The "Open in Browser" button will only show for remote repos

https://github.com/user-attachments/assets/fd0fed67-86f8-4a37-b580-50376d0bac6f


love-linger:
You can do this from the context menu of remotes.

L1cardo:
> You can do this from the context menu of remotes.

I know, but it is not convenient. 1 click vs 3 clicks
image

love-linger:
* `Open in browser` is a feature about remotes not local repository
* If you have multiple remotes, the first one may not be the target you want to visit it in browser.

L1cardo:
Then how about using a Context menu? Like this
image

love-linger:

Issue #1293, opened 2025-05-08
Switch to "Local Changes" tab when Rebase fails with conflicts


lvde0:
When a rebase fails with merge conflicts it should switch back to the "Local Changes" tab to allow the user to inspect & fix the conflicts

PR #1294, opened 2025-05-08
Switch to "Local Changes" tab when rebase fails with conflicts

lvde0:
Fixes https://github.com/sourcegit-scm/sourcegit/issues/1293

### DIFF
--- a/src/ViewModels/Rebase.cs
+++ b/src/ViewModels/Rebase.cs
@@ -51,8 +51,19 @@ namespace SourceGit.ViewModels
             return Task.Run(() =>
             {
                 new Commands.Rebase(_repo.FullPath, _revision, AutoStash).Use(log).Exec();
+
                 log.Complete();
-                CallUIThread(() => _repo.SetWatcherEnabled(true));
+                CallUIThread(() =>
+                {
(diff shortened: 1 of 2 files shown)


love-linger:
I don't quite like such changes.

* Is it necessary to immediately call `_repo.RefreshWorkingCopyChanges()` after the `rebase` is completed? If there are conflicts, after calling it, since the content of the working directory files has changed, it will be called again by filesystem watcher.
* This will lead to a lack of consistency in the behavior of `Rebase/Cherry-Pick/Merge/Revert` because we haven't dealt with similar situations for other commands. Even just for the `Rebase` operation, for example: when there is an empty commit (both branches have `cherry-picked` a certain commit), git allows modifying the commit message for the empty commit to explain, but we don't navigate to the `LOCAL CHANGES` page.
* Currently, when there are conflicts, not only will the `In-Progress` toolbar be displayed, but also a ⚠️ sign will appear on the right side of `LOCAL CHANGES` to remind users that they may need to switch to this page. I think this is sufficient. In most cases, I'd like the idea of "letting users know what they can do" more than "deciding what to do for users".

PR #1305, opened 2025-05-12
Separate icon for annotated tag (to distinguish from lightweight tag)

goran-w:
The QueryTags command now additionally reads the %(objecttype) field, which allows us to distinguish between Lightweight tags and Annotated tags.

A new icon is applied for Annotated tags, almost everywhere - except for Models.Decorator and Models.Filter, since these are not (yet) updated to discern the specific tag type. Nevertheless, this change makes it much easier to distinguish if a tag is Annotated or Lightweight.

Also, it's now easier to understand if the tooltip tag-message comes from the (Annotated) tag itself or from the (Lightweight) tag's corresponding commit.

Finally, a `Border` is added around the `TextBlock` of items in the TAGS list, to make its message-tooltip work more similar to the ref-tooltip in the LOCAL BRANCHES and REMOTES lists.


love-linger:
The commit 6a5e6d12d70f52e5777cc4edc4022fed870151d4 has been cherry-picked.

Regarding the icons for tags, I've decided to keep them consistent. Whether a tag is `annotated` or `lightweight`, its functionality remains unaffected.

Also, when retrieving the tag list, for `annotated` tags, we always prefer to point the `SHA` property to the referenced commit rather than the tag's own object. See line 35 of the code:
https://github.com/sourcegit-scm/sourcegit/blob/01625ada1a5d632b033519e2700516c971344840/src/Commands/QueryTags.cs#L32-L38

What `SourceGit` needs to do is to correctly display the tag's message and, upon click, correctly locate the referenced commit.

Most other git clients do not distinguish between these two types of tags either. Introducing different tag icons actually increases the user's cognitive cost (in fact, there's no place to inform users why there are two types of icons).

goran-w:
I see. Could we PLEASE then instead indicate, only in the tag-message tooltip itself, by an icon and/or by a prefix text like "[Annotated:]" and "[Log:]" (or similar) respectively, where the message comes from?

love-linger:
Done.

* `annotated` tag 
image
* `lightweight` tag
image

goran-w:
Thanks, @love-linger - looks great! Just a note: the "annotated" string will need to be localized...

goran-w:
Also, maybe omit the duplicated tag-name (on the message line) for a lightweight tag, if the commit-message is actually empty?
image

love-linger:
> Also, maybe omit the duplicated tag-name (on the message line) for a lightweight tag, if the commit-message is actually empty?

Can you test the output of git command in `QueryTags` for the `empty-commit`?

goran-w:
> Can you test the output of git command in `QueryTags` for the `empty-commit`?

Ah, Git itself seems to respond with the tag-name if the (lightweight) commit-message is empty, and even when the (annotated) tag-message is empty!

Thus, to avoid the duplication we would need to compare the stripped message to the tag-name and clear the message if they are identical...

goran-w:
Thank you, it now works as a charm!

And just for the record - my test result above was half wrong, but the fix is relevant nonetheless:
For a lightweight tag, an empty commit-message is returned as-is (empty) by Git.
(But for an annotated tag without a tag-message, Git will indeed return the tag-name, as I reported.)

Issue #1267, opened 2025-05-01
[Feature Request] Keyboard Shortcut(s) for switching Workspaces


alensiljak:
Now that you've spoiled us, I think switching workspaces could be a bit more efficient. A keyboard shortcut would save a few clicks and switching from keyboard to the mouse. 
For example
- Alt+1, 2, 3... to switch to a specific workspace. Works for the first N items.
- Ctrl+Alt+ <- / -> to switch sequentially through the list (up/down, or left/right)

I don't have specific preferences. Any of these would work. Perhaps you already had something in mind in this regard.
Also, perhaps using Alt directly for features that are not so frequently used is not the best option but that's up to you. It could also be Ctrl+Alt.

Actually, some common shortcuts for switching workspaces / desktops:
- Windows + Ctrl + Left / Right, to move through the desktops in Windows
- Windows + Tab, to open the list of desktops/workspaces in Windows
- Ctrl+Alt+ left/right, switches desktops in Gnome
- Windows + <number>, to switch to particular desktop in Gnome

However, obviously using the same one would conflict with the OS behaviour.

love-linger:
Done. You can download the latest CI build from Github Action.

Add hotkey `Ctrl+P/⌘+P` to open workspace switcher.

love-linger:
Sorry, in the latest commit, the hotkeys have been changed:

- change the hotkey to open `Preferences` window to `Ctrl+,/⌘+,`
- change the hotkey to open `Switch Workspace` popup to `Ctrl+Shift+P/⌘+⇧+P`
- change the hotkey to open `Switch Tab` popup to `Ctrl+P/⌘+P`

PR #1328, opened 2025-05-17
Added workspaces shortcuts

popara96:
- added Alt+Space for opening Workspaces context menu (which can then be navigated normally with arrows)
- added Alt+1 through Alt+9 for switching to corresponding workspace

This references #1267


love-linger:
I‘ll remove the hotkey to open workspaces drop-down menu. I do not like it at all. But the feature of switching workspaces is nice.

love-linger:
I've tested this feature. 

* The `⌥+1` and `⌥+2` are hard to trigger on macOS (others, such as `⌥+3`, are just work fine). I do not know why.
* It's not easy for users to remember their workspace's index if they have many workspaces.

So, I decided to rewrite this feature.

love-linger:
I've pushed a new implementation for workspace switching.

Users can use `Ctrl+P/⌘+P` to open workspace switcher.

PR #1330, opened 2025-05-17
Avoid duplication of Repository tab

Guddiny:
I and my colleagues have noticed that with a large number of open repositories it is much more convenient to find and navigate using a new tab. This tab contains a search with a list of already opened repositories, which significantly speeds up and simplifies work. In fact, we use this page as an analog of the `Go to everything` in Rider or `Code search` in VS (which is the same hotkey combination `Ctrl+T`).
And thus using this plug-in during the day you can create quite a few copies of it, which is not convenient.

In this PR it is suggested to keep a new tab in one copy at most.

### DIFF
--- a/src/ViewModels/Launcher.cs
+++ b/src/ViewModels/Launcher.cs
@@ -134,7 +134,7 @@ namespace SourceGit.ViewModels
         {
             var existingRepositoriesPAge =
                 Pages.FirstOrDefault(p => string.IsNullOrEmpty(p.Node.Name));
-            if (existingRepositoriesPAge != null)
+            if (existingRepositoriesPAge is not null)
             {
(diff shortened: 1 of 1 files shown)


love-linger:
It's a little wired when use click the `+` button in tab bar.

Guddiny:
Thank you for response. I don't feel it so strange. But your point is also valid.
What if you move an already opened task (if there is one) to the end of the list where it is expected to appear?

image

love-linger:
I've pushed a new implementation to switch tabs.

Users can now use `Ctrl+P/⌘+P` to open `Switch Tab` popup or `Ctrl+Shift+P/⌘+⇧+P` to open `Switch Workspace` popup

image

Guddiny:
Cool thank you!

PR #1351, opened 2025-05-22
fix: Combine branch and commit refresh into a single thread

gadfly3173:


goran-w:
Good catch! Please check also all the other occurrences of `Task.Run(RefreshCommits)` (in class `Repository`)...

Could the same issue occur for `RefreshCommits` vs `RefreshTags`? (If so, there's yet another case in class `Watcher`.)

gadfly3173:
> Could the same issue occur for `RefreshCommits` vs `RefreshTags`? (If so, there's yet another case in class `Watcher`.)

Currently, only merge/pull/fetch operations need to navigate to the head commit of the branch after refreshing the commit. Therefore, it is necessary to update the branch information first only when refreshing the branch and triggering the refresh through the watcher.

goran-w:
> Currently, only merge/pull/fetch operations need to navigate to the head commit of the branch after refreshing the commit.

I see - so there's no need to also patch `Repository.RefreshAll()` and `Repository.MarkBranchesDirtyManually()` where branches and commits are refreshed in parallel?

BTW, hopefully this will fix an issue I've noticed a few times, where a branch-badge is still shown at an older commit after an operation that should move it, and there even being two instances of the same branch-badge in the log-graph (one at the incorrect older commit and one at the correct newer one)!

love-linger:
I do not like this change. I've pushed a new commit c112549b69d74d6583bc4fe7cb1e97a516fc106e that query the changed branch head when we want to navigate it delayed (after commits have been refreshed)

PR #1355, opened 2025-05-23
localization: update spanish translations

jjesus-dev:
add missing translations

- Text.DeinitSubmodule.Force says: _Force de-init event if_...

@love-linger I need to ask, it is "_de-init event if_" or "_de-init even if_"? because that could change the translation (I took it like "**event if**").


love-linger:
Sorry, it is `de-initialize submodules even if` not `event if`

PR #1356, opened 2025-05-23
Fix IsConflictResolved check for submodule

goran-w:
Improved IsConflicResolved check, to work correctly for submodule conflicts.

~~Added an extra (lock-free) call to git-status with `--porcelain=v2` which makes it easy to check that:~~
* ~~The file is still "unmerged".~~
* ~~The file is a submodule (or not).~~

If the file is a submodule, we always ~~return `false` (while it's unmerged)~~ regard the conflict as unresolved (while Unstaged), since we can't use text-diff check on submodules and the conflict is not resolved until it's Staged.

### DIFF
--- a/src/ViewModels/Conflict.cs
+++ b/src/ViewModels/Conflict.cs
@@ -1,4 +1,6 @@
-﻿namespace SourceGit.ViewModels
+﻿using System;
+
+namespace SourceGit.ViewModels
 {
     public class ConflictSourceBranch
     {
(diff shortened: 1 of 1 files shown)


goran-w:
@love-linger I rewrote my commit, the new solution is much simpler (if the file is a submodule entry we just bypass IsCommitResolved and regard the conflict as unresolved - it will be resolved when it's Staged).

PR #1357, opened 2025-05-23
[Feature] Support for WSL git

roguesaloon:
Allows for sourcegit to use WSL git (from default distribution) when working inside a WSL directory with UNC path (i.e. `\\wsl.localhost`) on Windows. Dynamically chooses to use WSL git over windows git based on the filepath.

This allows users to manage WSL repos through the same windows instance of sourcegit alongside windows repos, which is a huge win for my own personal workflow. Partly based on similar WSL implementation in GitExtensions.

### DIFF
--- a/src/Commands/Command.cs
+++ b/src/Commands/Command.cs
@@ -167,6 +167,19 @@ namespace SourceGit.Commands
                 start.Environment.Add("LANG", "C");
                 start.Environment.Add("LC_ALL", "C");
             }
+            else if (OperatingSystem.IsWindows())
+            {
+                // Use WSL git for WSL paths on Windows
+                var wsl = new Models.WSL() { Path = WorkingDirectory };
+                if (wsl.IsWSLPath())
+                {
+                    start.FileName = "wsl";
+                    start.Arguments = $"git {start.Arguments}";
+
+                    wsl.SetEnvironmentForProcess(start);
+                    selfExecFile = start.Environment["SSH_ASKPASS"];
+                }
+            }
 
             // Force using this app as git editor.
             switch (Editor)
--- a/src/Commands/Command.cs
+++ b/src/Commands/Command.cs
@@ -31,6 +31,7 @@ namespace SourceGit.Commands
         public EditorType Editor { get; set; } = EditorType.CoreEditor;
         public string SSHKey { get; set; } = string.Empty;
         public string Args { get; set; } = string.Empty;
+        protected Models.WSL WSL { get; private set; }
 
         // Only used in `ExecAsync` mode.
         public CancellationToken CancellationToken { get; set; } = CancellationToken.None;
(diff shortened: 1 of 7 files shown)


Bluejanis:
Tested on current `develop` (Windows 11 + Ubuntu-24.04). The dispatch works well, but
SSH remotes fail when auth goes through an ssh-agent:

```
> wsl git ls-remote origin HEAD
git@github.com: Permission denied (publickey).
```

`SSH_AUTH_SOCK` is usually exported from a shell rc, and `wsl.exe` runs commands
non-interactively — stock Debian/Ubuntu `.bashrc` (from `/etc/skel`) returns early in
that case, so it is never set. `bash -lc` does not help either: `.profile` sources
`.bashrc`, which then bails. HTTPS remotes and on-disk keys are unaffected.

Fix, using the WSLENV forwarding already in `SetEnvironmentForProcess`: probe once for
a live agent socket — ask `$SHELL -ic` for `$SSH_AUTH_SOCK`, falling back to known
paths — then add `SSH_AUTH_SOCK` to `start.Environment` and to the `wslEnvirionment`
list. No agent means nothing is added, so setups that work today are untouched.

Separately, the `SetWatcherEnabled` hunk no longer applies — that method was removed
from `Repository.cs`, which is why this shows as conflicting. The other two hunks
cherry-pick onto `develop` cleanly without it.

Happy to send a PR against your branch if that is easier.

roguesaloon:
> Tested on current `develop` (Windows 11 + Ubuntu-24.04). The dispatch works well, but SSH remotes fail when auth goes through an ssh-agent:
> 
> ```
> > wsl git ls-remote origin HEAD
> git@github.com: Permission denied (publickey).
> ```
> 
> `SSH_AUTH_SOCK` is usually exported from a shell rc, and `wsl.exe` runs commands non-interactively — stock Debian/Ubuntu `.bashrc` (from `/etc/skel`) returns early in that case, so it is never set. `bash -lc` does not help either: `.profile` sources `.bashrc`, which then bails. HTTPS remotes and on-disk keys are unaffected.
> 
> Fix, using the WSLENV forwarding already in `SetEnvironmentForProcess`: probe once for a live agent socket — ask `$SHELL -ic` for `$SSH_AUTH_SOCK`, falling back to known paths — then add `SSH_AUTH_SOCK` to `start.Environment` and to the `wslEnvirionment` list. No agent means nothing is added, so setups that work today are untouched.
> 
> Separately, the `SetWatcherEnabled` hunk no longer applies — that method was removed from `Repository.cs`, which is why this shows as conflicting. The other two hunks cherry-pick onto `develop` cleanly without it.
> 
> Happy to send a PR against your branch if that is easier.

Hi @Bluejanis, Thank you for your interest in this. The original PR was quite primitive and not very complete from a functionality perspective (more of a POC really), and based from a much older version of sourcegit. I have actually been working on this as a side project for while now, and have worked out a lot of the edge cases and practical problems of using WSL git with a windows GUI. This version uses a persistent shell for frequently executed background commands, to stop constant process churn.

I have been using this build daily at work with WSL on Windows 11 and Debian 13 and it works well for me. I had been meaning to update this PR for a while, as the majority of this code hasn't changed in a while, but hadn't got round to it. So now I've pushed what I have. It is still based from an older develop branch and has some conflicts, but they should be easier to fix than before.

PR #1362, opened 2025-05-24
fix: disable squash/fixup actions for the first changeable commit

gadfly3173:


love-linger:
I've pushed my implementation of this feature. See commit 0e35c565299e4c83b75d0e94115fd50c26abd5f0

PR #1375, opened 2025-05-29
del BuiltInComInteropSupport>true as it contradicts the line below

M-L-Ml:


love-linger:
Confirmed. It does not need this attribute.

PR #1386, opened 2025-06-04
[Feature] Show Diff of LFS content

henrik-andersson-sus:
We have a lot of files in lfs storage, and it would be a nice to be able to see the content of the diff. I've added a preference toggle to the diff view to de-reference the lfs pointers and use the already existing diff views for images and textures.

### DIFF
--- a/src/ViewModels/DiffContext.cs
+++ b/src/ViewModels/DiffContext.cs
@@ -140,7 +140,7 @@ namespace SourceGit.ViewModels
                 var ignoreWS = Preferences.Instance.IgnoreWhitespaceChangesInDiff;
                 var showContentInLFSDiff = Preferences.Instance.ShowContentInLFSDiff;
                 var latest = new Commands.Diff(_repo, _option, numLines, ignoreWS).Result();
-                var info = new Info(_option, numLines, ignoreWS, latest);
+                var info = new Info(_option, numLines, ignoreWS, showContentInLFSDiff, latest);
                 if (_info != null && info.IsSame(_info))
                     return;
 
@@ -310,14 +310,16 @@ namespace SourceGit.ViewModels
             public string Argument { get; set; }
             public int UnifiedLines { get; set; }
             public bool IgnoreWhitespace { get; set; }
+            public bool ShowContentInLFSDiff { get; set; }
             public string OldHash { get; set; }
             public string NewHash { get; set; }
 
-            public Info(Models.DiffOption option, int unifiedLines, bool ignoreWhitespace, Models.DiffResult result)
+            public Info(Models.DiffOption option, int unifiedLines, bool ignoreWhitespace, bool showContentInFLSDiff, Models.DiffResult result)
             {
                 Argument = option.ToString();
                 UnifiedLines = unifiedLines;
                 IgnoreWhitespace = ignoreWhitespace;
+                ShowContentInLFSDiff = showContentInFLSDiff;
                 OldHash = result.OldHash;
                 NewHash = result.NewHash;
             }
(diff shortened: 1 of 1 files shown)


love-linger:
I do not like this implementation. I've pushed a new commit ed496a41fbaf8ac545c996029795daeb50eaee40.

love-linger:
The reasons:

* Since `SourceGit` only shows the file size change for ordinary binary files, and this information is also displayed in the `LFS` diff view, we only need to handle the case when the `LFS` object points to an image file.
* Reading `LFS` file content directly from `.git/lfs/objects` is unsafe because after the user executes `git lfs prune`, this directory does not contain all the `LFS` objects.
* Switching the display mode in `Preferences` is also not a good idea in my opinion.

love-linger:
In my implementation, if the LFS points to an image, there're two tabs: `LFS` and `IMAGE`.

image

henrik-andersson-sus:
Fair enough 👍

henrik-andersson-sus:
The new implementation does not seem to support diff views of text files stored in LFS. Should it be implemented as a separate LFSTextDiff class as well then?

love-linger:
I don't understand why the `LFS` is used to manage plain text files.

goran-w:
> I don't understand why the `LFS` is used to manage plain text files.

@love-linger This is sometimes done with (large) text files, for example large `.csv` text files or large Unity `.asset` (YAML) text files...

goran-w:
@love-linger The main feature missing from your implementation of LFS image diff (compared to the one by @henrik-andersson-sus) is the Preference toggle of preferred view, `LFS` (Pointer) vs `IMAGE` (Content) - so we don't have to manually switch that EVERY time for EVERY image file! Could you please add that?

love-linger:
How to quickly determine whether an lfs object points to a plain text file? For image files, this is very easy, because AvaloniaUI supports limited image file types, and we can judge according to the file extension.

It is not advisable to read the whole file to do that job when selecting lfs changes. Whether in terms of operating efficiency or memory usage. For example, for large model files, lighting baking data in the game, navigation mesh of large scale game world, etc.

At present, there has been an obvious delay in loading image changes pointed to by lfs.

goran-w:
> How to quickly determine whether an lfs object points to a plain text file? For image files, this is very easy, because AvaloniaUI supports limited image file types, and we can judge according to the file extension. It is not advisable to read the whole file to do that job when selecting lfs changes. Whether in terms of operating efficiency or memory usage.

What was used in this PR was to let `git diff` determine (just as for non-LFS files) whether the content is `Binary` or can produce a text-diff. However, for LFS content the `--no-index` flag had to be used, in order to diff file-content that Git has cached to disk from LFS, but as you pointed out we should first make sure the data is actually cached...

Anyway, at the moment we can live without LFS text-diff, and the LFS image-diff is already a great improvement.

However, we definitely need the option persisted! (As a Preference variable - but by that I don't mean that it should be controlled via the Preferences UI, only that our preferred `LFS` / `IMAGE` option should be remembered - just like the `Ignore All Whitespaces` toggle and several others.)

If (for example) it takes too long to load image-diffs, we'll simply switch back to LFS-view, but the most likely case is that we want to view the next LFS image-file using the same view as the previous one (having to switch views EVERY time will quickly become frustrating). Besides, looking at LFS pointer SHAs is not the most fun you can have in a work-day, anyway... 😂

goran-w:
Related commit eebadd67a1e4f0507af3fb045b76b116493c2ba8 adds persisting the LFS / IMAGE option choice. 👍

PR #1387, opened 2025-06-04
Several improvements for the Welcome/Repositories page

goran-w:
* Unify all file-path normalization, to use char-replace (instead of string-replace) and to trim any trailing slash.
  * Removing (if present) the trailing slash in paths is needed to avoid failing comparisons, since `DirectoryInfo.Fullname` (and `.FullPath`) will not "normalize" the presence/absence of a trailing slash, making any direct equality tests error-prone.
  * This fixes a bug in ScanRepositories.GetUnmanagedRepositories(), where not all Git repo folders were added.
* Refactor the sorting of the RepositoryNode tree, to use a single method instead of duplicated code.
  * Also make the sorting of the RepositoryNode tree case-insensitive, to make items easier to locate.
  * Sort & Save the whole RepositoryNode tree after a complete rescan, to make sure it's fully consistent.
* Activate TabsDropdownItem on Tapped instead of DoubleTapped, for easier access.
* Bug-fix: Prevent exception in Repository.RefreshBranches(), for repo which has no commits/branches yet.
* Correct spelling error in App.GetLauncher() method.
* Correct a variable name from 'founded' to 'found'.

### DIFF
--- a/src/ViewModels/Repository.cs
+++ b/src/ViewModels/Repository.cs
@@ -1120,7 +1120,8 @@ namespace SourceGit.ViewModels
                 if (_workingCopy != null)
                     _workingCopy.HasRemotes = remotes.Count > 0;
 
-                GetOwnerPage()?.ChangeDirtyState(Models.DirtyState.HasPendingPullOrPush, !CurrentBranch.TrackStatus.IsVisible);
+                var hasPendingPullOrPush = CurrentBranch?.TrackStatus.IsVisible ?? false;
+                GetOwnerPage()?.ChangeDirtyState(Models.DirtyState.HasPendingPullOrPush, !hasPendingPullOrPush);
             });
         }
 


love-linger:
I've manually cherry-picked partial of commits in your PR. 

Commit 663567e700f18926a2ce29581919e0385f6b180c is not picked. I think sort by name in case-sensitive way is more common than case-insensitive and I really do not like it. For example:

```
A
a
Ab
ac
ad
Ae
```

BTW. Next time, please don't submit so many commits (for different purpose) in a single PR. It's difficult to tell from GitHub which changes correspond to which commit. I can only cherry-pick each commit manually and then review. It's too tiring.

goran-w:
Thank you. I'll use fewer commits and more concise PR branches in the future. But also, you could've asked me and I would have removed the unwanted commit from my PR branch. 😉 

However, personally, I find case-sensitive sorting very frustrating, since `bee` is not located between `Beat` and `Bell` in the example below - I have to look through two "sections" of the list to find it...

```
Art
Beat
Bell
Boo
Cow
Deep
Early
Far
arty
bee
cee
dee
fee
```

So at least it should be made a global Preference "Use case-insensitive sorting in all lists" (or similar)! 

Let's close this PR, I'm working on a new one for optional case-insensitive sorting (including some bug-fixes). 😬

goran-w:
Follow-up PR for case-insensitive sorting Preference option: #1401

goran-w:
For reference, the commits cherrypicked from this PR were:
f04b0c5, 75c32c1, 54c05ac, 88c38b4, b969ac1, 406ace9
(Also, commit 464fe74 was a code review for b969ac1.)

PR #1395, opened 2025-06-05
[Feature] Added missed LFS Image cases to file browser and file history

henrik-andersson-sus:
I tried to follow in the style of the LFSDiffView. Possibly the function BitmapFromLFSObject could be moved somewhere more appropriate to be able to reuse it instead of changing its access modifier

### DIFF
--- a/src/ViewModels/CommitDetail.cs
+++ b/src/ViewModels/CommitDetail.cs
@@ -230,9 +230,23 @@ namespace SourceGit.ViewModels
                         var matchLFS = REG_LFS_FORMAT().Match(content);
                         if (matchLFS.Success)
                         {
-                            var obj = new Models.RevisionLFSObject() { Object = new Models.LFSObject() };
-                            obj.Object.Oid = matchLFS.Groups[1].Value;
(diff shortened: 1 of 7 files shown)


goran-w:
NOTE: This PR is a follow-up to previous PR #1386.

love-linger:
Sorry. I need to re-write all codes to detect lfs pointer and image loading.

goran-w:
@love-linger - Currently there's a discrepancy / inconsistency between the Diff and Files views. We should be able to view an LFS image File as IMAGE, just like we are (since recently) able to view an LFS image Diff as IMAGE.

Can you please explain the future of the suggested fix from this PR? Do you intend to implement it, but in a slightly different way? (If so, could you add an issue so we can track its progress?)

Also, what was bad about the suggested implementation?

love-linger:
I mean I've rewrite the codes about lfs pointers detection and image loading. And this request has been added in my code. See latest develop branch.

image

goran-w:
Ah, sorry - I missed that this was addressed in commit a023a9259bd8005110c97003db02d893478cc87b. (It's harder to follow updates when the related PR is not referenced in the commit.)

PR #1396, opened 2025-06-05
[Feature] Keep tree state and selection in 'Files' tab when changing commit in history

johanw1232:
In the 'Files' tab for commit history window when previewing the selected file you lose the selection and expand state of all the nodes when selecting a new commit. It would be nice to keep the selection (if possible) when jumping between commits, for a quick glace of the changes made in a single file.

This PR aims to keep the selected item and update the preview to the new commit.

### DIFF
--- a/src/Views/RevisionFileTreeView.axaml
+++ b/src/Views/RevisionFileTreeView.axaml
@@ -8,7 +8,7 @@
              mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
              x:Class="SourceGit.Views.RevisionFileTreeView"
              x:Name="ThisControl">
-  <v:RevisionFileRowsListBox ItemsSource="{Binding #ThisControl.Rows}"
+  <v:RevisionFileRowsListBox x:Name="RevisionFileRowsListBox" ItemsSource="{Binding #ThisControl.Rows}"
                              Background="Transparent"
                              SelectionMode="Single"
                              SelectionChanged="OnRowsSelectionChanged"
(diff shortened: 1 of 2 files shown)


love-linger:
* The directory structures may vary for different revisions.
* Expanding a folder tree node means running the command `git ls-tree -z <REVISION> -- <FOLDER_PATH>` once. This will seriously affect the response speed of the entire interface. If the hierarchy is too deep, it may even cause the interface to freeze.
* For the need to view the changes of the same file under different revisions, it is recommended to use the `File Histories` feature.

goran-w:
> * The directory structures may vary for different revisions.

That is taken care of quite gracefully in this PR, by simply losing the selection (while still keeping the parent folder(s) expanded, if possible).

> * Expanding a folder tree node means running the command `git ls-tree -z <REVISION> -- <FOLDER_PATH>` once. This will seriously affect the response speed of the entire interface. If the hierarchy is too deep, it may even cause the interface to freeze.

I don't see a big slowdown even several levels deep in a folder with lots of files. (If this is an issue, we could have an option "Preserve selection in file-tree" or similar.)

> * For the need to view the changes of the same file under different revisions, it is recommended to use the `File Histories` feature.

That is not really the same thing, since this feature also shows the selected file in context at its path in the repo.

Having to drill down repeatedly to the same subfolder hierarchy each time another Commit gets selected is simply frustrating, and it makes the `FILES` view less useful than it could be.

heartacker:
this is usefull

goran-w:
Rebased the branch, to work with the newer async handling. May need some extra work just to remove "flickering" when re-expanding file-tree levels (should be done all-at-once before re-displaying tree).

PR #1397, opened 2025-06-05
Typos

NathanBaulch:
Awesome project! Easily the most powerful Git GUI I've found on Linux (speaking as a previous gitfork user).
It's not much, but here are some typo fixes to show my appreciation.

FWIW, the top typos are:
* launcer (20)
* unchaged (13)
* hightlighted (11)
* prefered (12)
* actived (7)
* expaneded (6)
* thinkness (5)
* highligh (4)
* singletone (3)
* feture (2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ You can find the current translation status in [TRANSLATION.md](https://github.c
 
 **To use this tool, you need to install Git(>=2.25.1) first.**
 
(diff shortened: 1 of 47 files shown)


gadfly3173:
Directly modifying the property name in `src/Models/RepositorySettings.cs` will cause the modified configuration to be lost after the upgrade.

love-linger:
At present, two configurations are affected, `OnlyHighlightCurrentBranchInHistories` and `PreferredOpenAIService`. I don't think there is a big problem.

PR #1398, opened 2025-06-05
[UI] Unify looks between file viewer and history file content

henrik-andersson-sus:
Currently there is a ui discrepancy in the files content viewer, compared to the changes and file history content viewers, in that it is missing a header bar with it's filename. I've tried to unify the looks of them by moving the header from the FileHistories content viewer to a new intermediate FileContentViewer.

One of the reasons I think it is nice to unify is that it enables the option of adding toggle buttons such as the ones in the changes view when viewing content in the different places. As an example, I'm looking into creating a PR with the option to turn of bitmap interpolation, which currently makes small images very blurry and it would be nice to have a toggle mode in each of the image viewers.

Here is some images of the current ui:
Files browser without the header
image

Files browser with my proposed changes:
image

Changes tab and file history content with the header for reference:
image
image

My change does however introduce an empty field when no file is selected, which might be a personal preference if it's a bother or not.
image


love-linger:
I've pushed my implement of this PR

goran-w:
> I've pushed my implement of this PR

For reference, this was done in commit ac55bed.

Issue #597, opened 2024-10-24
Numerical Sorting in file listing


og-sims:
Hi,

in some projects the naming of files begins with numbers. On Windows system the default behavior ist sorting the files by string as follows:

```
T1
T10
T11
T2
T22
T25
T3
...
```

This behavier can be changed by GPO to a so called numeric sort:

```
T1
T2
T3
T10
T11
T22
T25
...
```

Here is an example file listing:

![image](https://github.com/user-attachments/assets/1b2f7e76-74c0-46f9-b587-63a3a5447014)

In larger Projects with 1000 and more objects, it is very hard to review code, because of scrolling up an down and searching numbers. 

Is there a way to change the sorting of file listings? When not, is this possible to implement this as option?

love-linger:
I've pushed a commit that enables numeric sorting in tree view. You can download the latest CI build from Github Action.

PS: For changes, you must change the display mode to `Show as Filesystem Tree` to use numeric sorting.

PR #1401, opened 2025-06-06
[feature] New Preference setting : Use case-insensitive sorting in lists

goran-w:
Whether to sort lists using case-sensitive or case-insensitive ordering is a (usually strong) matter of taste and conventions.

To satisfy both camps, this option can now be toggled as a (global) `Preferences` checkbox (on the `Appearance` tab). When toggling this option, a `RefreshAll()` is performed to directly update visible lists.

NOTE: This also affects the non-digit substrings in `NumericSort.Compare()`. Additionally, that method is now refactored to remove some redunant code.

Fixed a missing case of numeric sorting, for root items in the `RevisionFileTreeNode` tree (which were still using ordinal `string.Compare()` instead of the intended `NumericSort.Compare()`). This is related to issue #597 and commit 76a7a22 which added "numeric sorting for all trees". The issue is corrected by refactoring the node sorting into a separate method, to make sure sorting is done the same way in both of the call-sites.

To support sorting options in Local Changes (Unstaged / Staged) lists, the output list from `QueryLocalChanges()` is now sorted by `Path` (instead of relying "blindly" on the sorting done by Git). Also, `WorkingCopy.IsChanged()` is now simplified / optimized and will additionally react to changes in sorting order.

### DIFF
--- a/src/Commands/CompareRevisions.cs
+++ b/src/Commands/CompareRevisions.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Text.RegularExpressions;
+using SourceGit.ViewModels;
 
 namespace SourceGit.Commands
 {
@@ -39,7 +40,8 @@ namespace SourceGit.Commands
             foreach (var line in lines)
                 ParseLine(line);
 
-            _changes.Sort((l, r) => string.Compare(l.Path, r.Path, StringComparison.Ordinal));
+            _changes.Sort((l, r) => string.Compare(l.Path, r.Path,
+                Preferences.Instance.GetPreferredListComparisonType()));
             return _changes;
         }
 
--- a/src/Commands/QueryLocalChanges.cs
+++ b/src/Commands/QueryLocalChanges.cs
@@ -159,6 +159,8 @@ namespace SourceGit.Commands
                     outs.Add(change);
             }
 
+            outs.Sort((l, r) => string.Compare(l.Path, r.Path,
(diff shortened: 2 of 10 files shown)


goran-w:
@love-linger would you consider merging this PR? It's a bit frustrating to switch between SourceGit and other tools (like `Windows Explorer` 😬) that ignore case when sorting files... (SourceGit does a special "numerical sorting", for only certain lists, but does not support the much more basic sorting-option of optionally ignoring case.)

Personally, I find it harder to locate names in case-sensitive lists. For example, in the list below, `bee` is not located between `Beat` and `Bell` - I have to look through two "sections" of the list to find it...

```
Art
Beat
Bell
Boo
Cow
Deep
Early
Far
arty
bee
cee
dee
fee
```

(BTW, there's additionally a bug-fix for numerical sorting of root `RevisionFileTreeNode` items in the first commit of this PR.)

love-linger:
I've pushed a commit that use case-insensitive sorting in all filesystem related trees/lists by default. I do not want to add too many configurations.

goran-w:
@love-linger Thanks! However, I just realized that `Commit.ParseDecorators()` should also use the (case-insensitive) `NumericSort.Compare()` method, when comparing names for decorators (of same `Type`) in the call to `Decorators.Sort()`, since that's how `Branches` and `Tags` are sorted elsewhere...

love-linger:
Done

goran-w:
For reference, these are the related commits: 5494093, f59851f

PR #1407, opened 2025-06-09
Added button to disable image filtering

henrik-andersson-sus:
Added a button to disable image filtering. Currently bilinear interpolation is always enabled, which is a good default. However for small images as well as pixel-art images it would be a good to be able to toggle it on or off.

The icon for it is a subject for debate though 👍


love-linger:
I really do not want to add more and more options to users. I've pushed a new commit 69792b32622ef7726967a894ef8794cfcdef406e that never up-scales the image (but can down-scale if current space is smaller than the image's size)

goran-w:
For reference : commit c3c7d32 is also related.

PR #1408, opened 2025-06-09
[Feature] Update blame window when clicking/navigating commits

johanw1232:
This PR aims to update the blame window with the commit you click on in the left column.
It also adds a history so you can go back and forward through commits with clickable buttons and Mouse4/5 for back and forward.


love-linger:
very nice implementation. I will change `_shaHistory` to `_navigationHistory`. This name puzzled me for a while. I thought it was the list of revisions referenced in the entire `Blame`.

PR #1409, opened 2025-06-10
feature: double tap commit with tracked remote branch checks out loca…

NathanBaulch:
My first feature PR 🤞

Every so often I double click on an upstream branch tracked by a local branch that's a few commits behind. In this situation sourcegit prompts me to create a new branch, but it seems pretty rare that the user would actually want this.

Instead it should behave like double clicking on a remote branch in the branch tree - detect that a local tracking branch exists and check that out instead. This PR implements this behavior.

Might be even better to ask the user if they'd also like to fast-forward, but I'm keeping my first PR simple.

### DIFF
--- a/src/ViewModels/Histories.cs
+++ b/src/ViewModels/Histories.cs
@@ -231,9 +231,24 @@ namespace SourceGit.ViewModels
                         return;
                     }
                 }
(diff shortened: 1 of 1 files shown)


NathanBaulch:
Awesome, thanks for implementing "checkout & fast-forward"!

PR #1416, opened 2025-06-13
feature: double tap specific branch

NathanBaulch:
Checkout the branch the user actually double tapped on, not just the first branch decorator on the commit.

I am not experienced with Avalonia so this is definitely not an idiomatic implementation, but it works as a proof of concept. Most notably the `DecoratorAt` method I added to `CommitRefsPresenter` which maps the mouse location to a commit decorator based on where it was likely rendered.

Please refactor as needed!

### DIFF
--- a/src/ViewModels/Histories.cs
+++ b/src/ViewModels/Histories.cs
@@ -237,18 +237,18 @@ namespace SourceGit.ViewModels
                         {
                             if (localBranch.IsCurrent)
                                 return;
-                            if (localBranch.TrackStatus.Behind.Count == 0)
-                                _repo.CheckoutBranch(localBranch);
-                            else if (localBranch.TrackStatus.Ahead.Count == 0)
+                            if (localBranch.TrackStatus.Ahead.Count > 0)
+                            {
+                                if (_repo.CanCreatePopup())
+                                    _repo.ShowPopup(new CreateBranch(_repo, remoteBranch));
+                            }
+                            else if (localBranch.TrackStatus.Behind.Count > 0)
                             {
                                 if (_repo.CanCreatePopup())
                                     _repo.ShowPopup(new CheckoutAndFastForward(_repo, localBranch, remoteBranch));
                             }
                             else
-                            {
-                                if (_repo.CanCreatePopup())
-                                    _repo.ShowPopup(new CreateBranch(_repo, remoteBranch));
-                            }
+                                _repo.CheckoutBranch(localBranch);
                             return;
                         }
                     }


love-linger:
Double-clicking commit in HISTORY commit list to switch branches only provides a convenient way. In this mode, there are its unique rules, such as:

* User can checkout a commit as detached
* User can create a branch based on selected commit
* User can checkout the first local branch that referenced this commit

If the current commit is the HEAD of multiple local branches, the program also provides other ways, such as:

* User can checkout the target branch via double-clicking it from branch tree in left side-bar
* User can use the context menu entry `Checkout <branch>` of selected branch in branch tree
* User can use the context menu entry `Checkout <branch>` of selected commit in commit list (HISTORY)

For the functions of branches, in different parts of the software, the functions we provide are not exactly the same, and each has its own pertinence. For example, the functions related to branches in the context menu of the commit list are much fewer than those provided in the branch tree.

**Context menu of branch tree in left-sidebar**
image

**Context menu of selected commit in HISTORY**
image

NathanBaulch:
Understood. This PR makes double clicking on a commit ref `RenderItem` in history exactly the same as double clicking on a local or remote branch in the left sidebar.

love-linger:
What I mean is that I only hope the functions here to be simple and the rules to be clear.

In your implementation, double-clicking on a commit has too many different results.

* Double-clicking on an area that is not a refs will lead to the three situations I mentioned above.
* Double-clicking when the decorator is a local branch will switch to the specified local branch.
* Double-clicking when the decorator is a remote branch might has follow results
  - Create a new branch
  - Switch to its corresponding local branch (at this time, both might point to this commit, that is, this operation is equivalent to clicking on the local branch decorator in front of the one we are clicking)
  - Switch to its corresponding local branch and perform a fast-forward operation.
* Double-clicking when the decorator is a tag is equivalent to the first situation.

love-linger:
One more important thing is - The result of double-clicking on a commit must be strongly related to that commit.

In your PR: 

```csharp
if (localBranch.TrackStatus.Behind.Count == 0)
     _repo.CheckoutBranch(localBranch);
```

After checking out the local branch that tracking on the remote branch, the HEAD may be different with clicking one

You can review the original code of `DoubleTapped`. In the previous implementation, for all results, it was ensured that after switching branches, the HEAD of the current branch pointed to the commit we just double-clicked.

goran-w:
Minor side-note: if (double-/right-)clicking on a specific `Decorator` is supposed to have special meaning, then IMHO the visual `Decorator` itself should have hover-highlighting to indicate this possibility.

NathanBaulch:
I've tweaked the behavior so that double clicking on a branch ref/decorator in history exactly mirrors double clicking on that same branch in the left sidebar.
In short - double clicking on a remote branch with a local tracking branch that is ahead will now prompt to create a new local branch at that commit.

FWIW, I modeled this PR on the behavior of gitfork, which is what I personally migrated from (when I used to be on Windows) and what this project seems to take some inspiration from.

Issue #1454, opened 2025-06-24
Rescan Repositories should be aware of Workspace Default Clone Dir


henry-js:
I believe there should be a general preferences bool to "Prefer Workspace Default Clone Dir"
- When enabled, Rescan Repositories should take the current workspace clone dir instead of the default clone dir
- Rescan Repositories should clear repository nodes before scanning. This will remove invalid entries

goran-w:
I agree.

Workspaces don't use their configured Default Clone Dir as expected.
* A Workspace has its own Default Clone Dir, but the one used when scanning is (only) the global (redundant?) one! Instead, the global one could be used only if the Workspace has not configured a Default Clone Dir of its own.
* The Repositories list is the same, regardless of which Workspace we choose! Instead, it should start empty for a new Workspace.

BTW : the concept of (hierarchical) Groups in this view is slightly confusing, since Groups are auto-created from directories when (re-)scanning the Default Clone Dir but they can also be re-arranged freely (the latter makes them diverge from the "physical" folder-structure on disk).

love-linger:
> The Repositories list is the same, regardless of which Workspace we choose! Instead, it should start empty for a new Workspace.

Completely disagree with this view. The design purpose of Workspace in SourceGit is to group the opened repositories and quickly switch between any Workspaces.

You can add any repo (open another repo that not added to current workspace) to current workspace at any time. There's no need to manage different repo list.

goran-w:
> Completely disagree with this view. The design purpose of Workspace in SourceGit is to group the opened repositories and quickly switch between any Workspaces.

Indeed, but for that to be really useful, I think the Workspaces should have their own separate lists of repositories?

> You can add any repo (open another repo that not added to current workspace) to current workspace at any time. There's no need to manage different repo list.

The issue here is that when creating a new Workspace, it already has ALL the repos from the earlier/previous workspace in its repositories-list. So when switching between Workspaces, there's no difference seen on the Welcome page...

(This also makes the scanning of Default Clone Dir feature less useful. Consider two separate Workspaces configured with two separate Default Dirs - scanning in each Workspace would then make repos from both Dirs appear in the repositories-list for both Workspaces.)

BTW, if a lot of repos are added to the repositories-list, it becomes cumbersome to clear the list - since multi-selection is not possible, we'd have to remove each item separately?

goran-w:
[...]

PR #1456, opened 2025-06-24
wip: add "prefer workspace clone dir" to preferences

henry-js:
This begins to address #1454
- Have added a new bool prop for PreferWorkspaceDefaultCloneDir to Preferences
- Added to Preferences UI

I am not sure what the ideal way to clear already existing ManagedRepositories from the Welcome ViewModel, so any guidance would be appreciated.

### DIFF
--- a/src/Resources/Locales/en_US.axaml
+++ b/src/Resources/Locales/en_US.axaml
@@ -514,6 +514,7 @@
   <x:String x:Key="Text.Preferences.General.ShowChildren" xml:space="preserve">Show children in the commit details</x:String>
   <x:String x:Key="Text.Preferences.General.ShowTagsInGraph" xml:space="preserve">Show tags in commit graph</x:String>
   <x:String x:Key="Text.Preferences.General.SubjectGuideLength" xml:space="preserve">Subject Guide Length</x:String>
+  <x:String x:Key="Text.Preferences.General.PreferWorkspaceDefaultCloneDir" xml:space="preserve">Prefer Workspace Default Clone Dir</x:String>
   <x:String x:Key="Text.Preferences.Git" xml:space="preserve">GIT</x:String>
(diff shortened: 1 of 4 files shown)


goran-w:
Alternatively, the "global" Default Clone Dir could be used if the Workspace has not already configured its own Default Clone Dir (in the Configure Workspaces dialog). Then we don't need the new option in Preferences...

Also, each Workspace should start "empty", instead of inheriting all the added repos from the earlier ones.

jamesshenry:
Ok I can do point 1,
For point 2, I understand each WS starts empty but when switching back and forth, the list of scanned repos persists becuse it is saved to `preference.json`. So unless the user manually clears and rescans every time, the Welcome screen will display a combination of all WS repos scanned

goran-w:
> Ok I can do point 1,

I think @love-linger already addressed this, in two recent commits?

> For point 2, I understand each WS starts empty but when switching back and forth, the list of scanned repos persists becuse it is saved to `preference.json`.

Yes. To further clarify: each new Workspace starts empty with regards to its open repo-tabs, but the repositories-list displayed on the Welcome page is not emptied and is shared between all Workspaces.

> So unless the user manually clears and rescans every time, the Welcome screen will display a combination of all WS repos scanned

Yes. There is currently no quick way to clear the repositories-list (I think).

Also, I'd prefer if the repositories-list was NOT shared between all Workspaces. For two Workspaces with two different values for `Default Clone Dir`, it would make more sense if scanning for repositories in both of these Workspaces would not create a mix of both these Default Dirs in the repositories-list for both Workspaces...

goran-w:
@henry-js I think this PR can be closed without merging, since the initial issue has been resolved elsewhere (see issue #1454)?

(NOTE: There are other remaining (follow-up) enhancement requests in the lengthy discussion under that issue, where PRs could be made...)

jamesshenry:
OK awesome, I'll have a look

PR #1458, opened 2025-06-25
Repo-search: apply filter also on Repos under matching Group-name

goran-w:
Addresses part of the discussion in issue #1454.

Group and Repo nodes are now name-filtered the same way, while still recursing into Groups with matching sub-Nodes.

It makes filtering more effective, instead of favoring Group-matches over Item-matches.


goran-w:
It was concluded in #1454 that this PR is not necessarily an improvement...

PR #1461, opened 2025-06-25
localization: update German translations

sina-hide:
It's translation time again.  ;-)

@love-linger You changed the translation of key `Text.Configure.CustomAction.Arguments.Tip` to reflect the selected tag as additional possibility for an argument.  Since this key was only modified and not new, it was not mentioned in "TRANSLATION.md".  It would be nice if such keys could be tagged automatically in any way, so that the translation can be checked and maybe adapted.  If you know of a way to accomplish this, it would be nice.


love-linger:
I'll remove the mismatched translations

sina-hide:
> I'll remove the mismatched translations

That's a pragmatic solution, and it's easy to do.  Nice.

PR #1481, opened 2025-07-01
Stream files

NathanBaulch:
I noticed a few places where files were being generated with `StringBuilder` in memory before writing to the FS.
Better to use `StreamWriter` and write directly to the FS rather than consuming unnecessary memory.
It's also a bit more readable and less error prone using `WriteLine` instead of `\n` IMO.
I've left the string interpolation in place because the code reads a little better. In theory those could be broken into multiple calls to `Write` to avoid even more string building behind the scenes.


src/Models/DiffResult.cs
```diff
@@ -163,40 +163,40 @@ public void GenerateNewPatchFromSelection(Change change, string fileBlobGuid, Te
...
                 var totalLines = Lines.Count - 1;
-                builder.Append("@@ -0,").Append(totalLines - additions).Append(" +0,").Append(totalLines).Append(" @@");
+                writer.WriteLine($"@@ -0,{totalLines - additions} +0,{totalLines} @@");
                 for (int i = 1; i <= totalLines; i++)
                 {
                     var line = Lines[i];
                     if (line.Type != TextDiffLineType.Added)
                         continue;
-                    builder.Append(selection.IsInRange(i) ? "\n+" : "\n ").Append(line.Content);
+                    writer.WriteLine(line.Content);
```
love-linger, line 172: We should not revert contents not in selected-range

PR #1487, opened 2025-07-01
ci:  code format check

NathanBaulch:
I noticed that trailing whitespace often slips through and has to be fixed in a `code_style` commit.
This PR adds a GH actions workflow that runs `dotnet format` on push. This check will fail on trailing whitespace or any other code formatting violation.


love-linger:
Why not let it format codes and commit if files are changed directly?

NathanBaulch:
Sure, if you like, I was just keeping it simple.
I personally prefer the original author to know that they're not following guidelines and hopefully adjust their future commits.

Issue #1439, opened 2025-06-20
Modernize async processing


NathanBaulch:
This project is full of `Task.Run` (157 calls) and `Dispatcher.UIThread.Invoke` (125 calls) to keep the UI responsive during I/O-bound operations (primarily waiting for `git` shell commands). As far as I can tell there's nothing CPU-bound in this app.

Rather than delegating I/O work to the thread pool and manually dispatching back to the UI thread, it's much better to take advantage of async/await for better responsiveness, lower overheads and more readable/linear code. More info [here](https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/async-scenarios).

I've started working on this but it's quite a significant refactor and wanted to check that this is something of interest before putting too much time into it. Is now a good time to adopt modern async patterns?

love-linger:
<s>I will try to modify a version first and test its efficiency compared with the current one. Before that, please do not submit the relevant PR.</s>

NathanBaulch:
Sure. It's not just about efficiency though, it's also about code readability (unless you're also referring to development efficiency).

Two random examples:

```csharp
// BEFORE
private void Gen()
{
    Text = string.Empty;
    IsGenerating = true;
    _cancel = new CancellationTokenSource();
    Task.Run(() =>
    {
        new Commands.GenerateCommitMessage(_service, _repo.FullPath, _changes, _cancel.Token, message =>
        {
            Dispatcher.UIThread.Invoke(() => Text = message);
        }).Exec();
        Dispatcher.UIThread.Invoke(() => IsGenerating = false);
    }, _cancel.Token);
}
// AFTER - able to use try-finally
private async Task GenAsync()
{
    Text = string.Empty;
    IsGenerating = true;
    try
    {
        _cancel = new CancellationTokenSource();
        await new Commands.GenerateCommitMessage(_service, _repo.FullPath, _changes, _cancel.Token, 
            message => Text = message).ExecAsync();
    }
    finally
    {
        IsGenerating = false;
    }
}
```

```csharp
// BEFORE
public override Task<bool> Sure()
{
    _repo.SetWatcherEnabled(false);
    ProgressDescription = "Archiving ...";
    var log = _repo.CreateLog("Archive");
    Use(log);
    return Task.Run(() =>
    {
        var succ = new Commands.Archive(_repo.FullPath, _revision, _saveFile).Use(log).Exec();
        log.Complete();
        CallUIThread(() =>
        {
[...]

PR #1489, opened 2025-07-02
refactor: auto dispatch notifications to UI thread in `Launcher`

NathanBaulch:
As part of my work on #1439 I'd like to remove any UI concerns from the `SourceGit.Commands` namespace.
I'm no Avalonia expert but it seems simple enough to detect the wrong thread in `Launcher` when processing notifications and dispatch to the UI thread as needed.
The only outstanding question is whether to use `UIThread.Post` or `UIThread.Invoke` as the existing code used a mixture of both.

### DIFF
--- a/src/Commands/Command.cs
+++ b/src/Commands/Command.cs
@@ -5,8 +5,6 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 
-using Avalonia.Threading;
-
 namespace SourceGit.Commands
 {
     public partial class Command
@@ -68,7 +66,7 @@ namespace SourceGit.Commands
             catch (Exception e)
             {
                 if (RaiseError)
-                    Dispatcher.UIThread.Post(() => App.RaiseException(Context, e.Message));
+                    App.RaiseException(Context, e.Message);
 
                 Log?.AppendLine(string.Empty);
                 return false;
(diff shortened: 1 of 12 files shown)


src/ViewModels/Fetch.cs
```diff
@@ -52,6 +52,8 @@ public Fetch(Repository repo, Models.Remote preferredRemote = null)
             {
                 SelectedRemote = _repo.Remotes[0];
             }
+
+            App.RaiseException("foo", "bar");
```
love-linger, line 56: Please remove the code for testing-only

NathanBaulch: Ha, nothing gets past you! Love it...

src/ViewModels/Archive.cs
```diff
@@ -57,12 +57,9 @@ public override Task<bool> Sure()
                 var succ = new Commands.Archive(_repo.FullPath, _revision, _saveFile).Use(log).Exec();
                 log.Complete();
 
-                CallUIThread(() =>
-                {
-                    _repo.SetWatcherEnabled(true);
-                    if (succ)
```
love-linger, line 63: I'd like do not move these code from `UIThread` to task's thread, since it is unnecessary to call `UIThread.Invoke` twice.

Issue #1156, opened 2025-04-04
Remember (or preset) the merge-flags option (in combo-box)


goran-w:
For our current workflow, in the `Merge Branch` dialog we use the `Merge Option` combo-box value `Don't commit` (`--no-ff --no-commit`).

We want to avoid having to change the combo-box setting on every `Merge`. Either of the following ways to achieve this would be fine:
* Remembering (even between sessions) the most recent choice made in the combo-box. (Possibly per repo.)
* A new combo-box option `Preferred Merge Option` (or similar) under the `Git` tab in `Preferences`. (NOTE: This should be per-repo as well!)

love-linger:
Since the merge operation is related to the target branch, for example, **some projects may restrict the merge rules for certain branches**, so `SourceGit` reads the git's `branch.<name>.mergeoptions` setting.

love-linger:
See #540

goran-w:
Makes sense, but AFAIK there's no "global" `mergeoptions` so for branches that don't have the config we'd need some way to specify/remember the combo-box choice.

love-linger:
![Image](https://github.com/user-attachments/assets/72c78759-0e3c-4b28-854c-8956f299fc8e)

Issue #540, opened 2024-10-03
default settings for merge options should be taken from .gitconfig


wlazlomarek:
Hi,
Whenever I want to merge two branches, the default options for that is:
![image](https://github.com/user-attachments/assets/16883829-d7d8-4623-a740-c20e465303ed)

So, I have to switch to `no fast-forward ` option each time.

I've checked if that may be control from .gitconfig but it is not. Any chance for some control? :)

love-linger:
now the default merge option will respect the `branch.<name>.mergeoptions` configuration. you can try the latest build from Github Action.

wlazlomarek:
Big thanks for the update, I will test this :)

PR #1495, opened 2025-07-02
enhance: respect git config `merge.ff`

sina-hide:
This adds another configuration possibility for the default merge mode to the already existing per branch git configuration `branch.<name>.mergeoptions` (#540) and the per repo preference "Preferred Merge Mode" (#1156).

Defined values:
 - `false`: do no fast-forward merge
 - `only`: do only a fast-forward merge

This is configurable by for example `git config set --global merge.ff <value>` for a global (meaning the current user) configuration.

The priority between these configurations/preferences is:

 - git configuration `branch.<name>.mergeoptions` (#540)
 - git configuration `merge.ff`
 - preference "Preferred Merge Mode" (#1156)

### DIFF
--- a/src/Models/MergeMode.cs
+++ b/src/Models/MergeMode.cs
@@ -2,13 +2,28 @@
 {
     public class MergeMode
     {
+        public static readonly MergeMode Default =
+            new MergeMode("Default", "Fast-forward if possible", "");
+
+        public static readonly MergeMode FastForward =
+            new MergeMode("Fast-forward", "Refuse to merge when fast-forward is not possible", "--ff-only");
+
+        public static readonly MergeMode NoFastForward =
+            new MergeMode("No Fast-forward", "Always create a merge commit", "--no-ff");
+
+        public static readonly MergeMode Squash =
+            new MergeMode("Squash", "Squash merge", "--squash");
+
+        public static readonly MergeMode DontCommit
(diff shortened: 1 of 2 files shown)


sina-hide:
@love-linger In addition to adding this feature I have done a few refactorings regarding `MergeMode`.  If you don't like them, I can revert them.

love-linger:
I think the `Preferred Merge Mode` in repository configuration (which can be traited as local setting) should use higher-priority than git config `merge.ff`.

I'll merge this first and modify it myself.

sina-hide:
> I think the `Preferred Merge Mode` in repository configuration (which can be traited as local setting) should use higher-priority than git config `merge.ff`.

Okay!

> I'll merge this first and modify it myself.

There was a little problem so `merge.ff` didn't work.  See PR #1501.

love-linger:
I just realized that when the user selects the `Default` merge mode, `SourceGit` does not pass parameters related to the merge mode to the `git merge` command. That is to say, if the user selects the `Default` merge mode, the `merge.ff` configuration should take effect automatically. We don't need the `GetGitConfigMergeFF` function.

Moreover, the description of the `Default` mode may be a bit misleading. It should not be `Fast-forward if possible`, but rather `Use git configuration`.

sina-hide:
You are right, @love-linger.  Your corresponding commit is d3d1377334d1bf8559686f3af09f328aa53bf652.

PR #1497, opened 2025-07-02
code_style: general cleanup

NathanBaulch:
Another batch of general cleanup.

### DIFF
--- a/src/ViewModels/CheckoutAndFastForward.cs
+++ b/src/ViewModels/CheckoutAndFastForward.cs
@@ -49,7 +49,7 @@ namespace SourceGit.ViewModels
             var updateSubmodules = IsRecurseSubmoduleVisible && RecurseSubmodules;
             return Task.Run(() =>
             {
-                var succ = false;
+                bool succ;
                 var needPopStash = false;
 
                 if (!_repo.ConfirmCheckoutBranch())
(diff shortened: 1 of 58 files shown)


src/ViewModels/CheckoutAndFastForward.cs
```diff
@@ -49,7 +49,7 @@ public override Task<bool> Sure()
             var updateSubmodules = IsRecurseSubmoduleVisible && RecurseSubmodules;
             return Task.Run(() =>
             {
-                var succ = false;
```
love-linger, line 52: I'd like to keep the orignal one, since the following line is also a boolean definition

src/Views/Histories.axaml
```diff
@@ -47,10 +47,6 @@
                 ContextRequested="OnCommitListContextRequested"
                 DoubleTapped="OnCommitListDoubleTapped"
                 KeyDown="OnCommitListKeyDown">
-        <DataGrid.Resources>
```
love-linger, line 50: Why remove this style? It overrides the default resource `DataGridGridLinesBrush` defined in `Avalonia.Controls.DataGrid`

src/ViewModels/CheckoutAndFastForward.cs
NathanBaulch: Sure, I only removed it because the `false` value is never used.

PR #1498, opened 2025-07-03
refactor: move I/O operations out of constructors

NathanBaulch:
This is important because constructors cannot be made async. This PR moves I/O operations into static `Create` factory methods or instance `Load` methods that are called from `OnLoaded` view events.

### DIFF
--- a/src/Commands/Command.cs
+++ b/src/Commands/Command.cs
@@ -32,7 +32,7 @@ namespace SourceGit.Commands
         public bool RaiseError { get; set; } = true;
         public Models.ICommandLog Log { get; set; } = null;
 
-        public bool Exec()
+        public virtual bool Exec()
         {
             Log?.AppendLine($"$ git {Args}\n");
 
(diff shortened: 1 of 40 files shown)


love-linger:
This PR will not be merged because I'm refactoring these code in feature/async

NathanBaulch:
Sure. For what it's worth, this PR is a good transitional step towards async without actually introducing any async (which is why I pushed it separately).

love-linger:
`OnLoad` will be triggered at every time this control is added to visual tree. For example: swithing between `HISTORIES` and `LOCAL CHANGES` page will trigger the conflict control multi-times.

NathanBaulch:
Fair enough, I admit I'm not familiar with the lifecycle of UI components. Hopefully there's a more appropriate place to put this code outside of view model constructors.

love-linger:
I completed the first refactoring of the `feature/async` branch.

I think we still need `Tash.Run` in constructors.

love-linger:
I don't think this modification is necessary. The current code is already quite concise.

PR #1499, opened 2025-07-03
refactor: just use `Wait` instead of `Stopwatch`

NathanBaulch:
No need for the `Stopwatch`.


love-linger:
This code has been included in `feature/async` branch

PR #1501, opened 2025-07-03
fix: `merge.ff` is ignored

sina-hide:
After changing the priority of the preferences/configurations for the merge mode in #1495, the git configuration `merge.ff` was ignored.  The reason is, that `GetSettingsPreferredMergeMode()` does not return `null` in the default case.  This is now fixed.


sina-hide:
Argh, my bugfix was buggy itself.  I should have tested it.

PR #1509, opened 2025-07-04
enhance: close all windows with `ESC`

NathanBaulch:
My muscle memory expects dialogs to be easily dismissed with `ESC`, so it's quite jarring when it doesn't work.
This PR ensures all chromeless windows close when escape is pressed.

### DIFF
--- a/src/Views/ChromelessWindow.cs
+++ b/src/Views/ChromelessWindow.cs
@@ -73,5 +73,13 @@ namespace SourceGit.Views
             if (sender is Border { Tag: WindowEdge edge } && CanResize)
                 BeginResizeDrag(edge, e);
         }
+
+        protected override void OnKeyDown(KeyEventArgs e)
+        {
+            base.OnKeyDown(e);
+
+            if (e is { Handled: false, Key: Key.Escape })
+                Close();
+        }
     }
 }
--- a/src/Views/Hotkeys.axaml.cs
+++ b/src/Views/Hotkeys.axaml.cs
@@ -1,5 +1,3 @@
-using Avalonia.Input;
-
 namespace SourceGit.Views
 {
     public partial class Hotkeys : ChromelessWindow
@@ -7,14 +5,6 @@ namespace SourceGit.Views
         public Hotkeys()
         {
             InitializeComponent();
-        }
-
-        protected override void OnKeyDown(KeyEventArgs e)
-        {
-            base.OnKeyDown(e);
-
-            if (!e.Handled && e.Key == Key.Escape)
-                Close();
         }
     }
 }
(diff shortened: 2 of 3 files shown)


love-linger:
I do not like this PR. `ESC` should only close toolbox windows (normally a modal dialog). Windows like `File Histories`, `Blame`, `Branch Compare` and `Dir Histories` should not use `ESC` to close.

NathanBaulch:
AFAIK none of these windows have unsaved state that the user needs to commit, so what's the risk of providing a quick dismiss option? Nothing is lost if the user accidentally hits `ESC`.

love-linger:
This has nothing to do with the content displayed within the window, but only with the window's own displaying mode. 

Usually, if a window is independent of another window, has `Minimize` and `Maximize` buttons, and has `Actived` and `Deactived` states, the `ESC` key will not be used to close the window. 

For example: on macOS, we use system caption buttons instead of the custom `CaptionButtons` view. When pressed the green button, the window will be changed to full-screen mode. And then user restore it using `ESC` key.

NathanBaulch:
I've never used macOS and have never seen `ESC` used to restore a maximized window in Linux or Windows, so perhaps that's the reason for my disconnect here.

goran-w:
In that case, let's use ESC for these windows on Windows/Linux but not on Mac. I'm on Windows and am also very much used to these kinds of programs conveniently using ESC in all "spawned" (and stateless) windows except for the very top-level one (i.e the SourceGit main window).

love-linger:
Have you really encountered windows that can be closed using the `ESC` key when it matches following rules:

- Has `minimized`, `maximized` and `restore` buttons
- Can be resized
- Has `Actived` and `Deactived` states

gadfly3173:
Wireshark’s Preferences window adheres to this rule. However, even if you minimize this window, the main window remains disabled until the Preferences window is closed.  However, it's more common to treat such pop-up windows like blame, history, etc., as dialog boxes—with only a close button—while still being standalone windows.

image

love-linger:
It is because the preference window of Wireshark is still a modal dialog.

Currently, for modal dialogs in SourceGit, users can close it by ESC key.

The reason that the blame and histories windows are not dialogs is:

- those windows are resizable
- When these windows exist, the main window must be able to operate, such as when you click some sha

goran-w:
The `WinMerge` tool offers several modes for this, in a single combo-box. As a frequent user of that tool, I find this flexibility awesome - it even allows (optionally) closing the main window with ESC. (It's using an MDI model, but the MDI windows are tileable & resizable and have min/max-imize and close buttons.)

image

image

NOTE: There's no real need to handle this differently in SourceGit on Mac - instead, this feature could be optional (in global Preferences) and disabled by default!

love-linger:
I no longer want to discuss this topic in depth because I have clarified the design principles of this software.

PR #1511, opened 2025-07-04
refactor: remove anonymous function in `ShowWindow`

NathanBaulch:
Just a suggestion, but I find nested anonymous functions a little hard to read. This PR moves the one an only instance of this in `ShowWindow` to a private method. There are no closures, so a straightforward change.


NathanBaulch:
No longer relevant with recent window refactoring.

Issue #977, opened 2025-02-12
[Feature Request] Discarded changes should be saved in a dated file in the Trash


dsteinbe-aip:
GitHub Desktop can save discarded changes in the Trash, this would be a nice feature. 😁

[Discarded changes are saved in a dated file in the Trash](https://docs.github.com/en/desktop/making-changes-in-a-branch/committing-and-reviewing-changes-to-your-project-in-github-desktop#discarding-changes)

love-linger:
Why not use `Stash` if you want to keep worktree clean and up to date to `HEAD`, and also save the changes before to another place?

SebastianSchumann:
I've the same requirement. Switching branches asks how to deal with changes. The last choice is restored and if you're a little bit off you'll discard your changes.

The discarding dialog should be enhanced by an option to store that changed to trash or not.

love-linger:
> I've the same requirement. Switching branches asks how to deal with changes. The last choice is restored and if you're a little bit off you'll discard your changes.
> 
> The discarding dialog should be enhanced by an option to store that changed to trash or not.

This is why I didn't want to add the feature requested in #185 in the first place...

mithom:
adding the option to store will only make the dialog more verbose/confusing. With still the same issue of being slightly off resulting in a discard. The current approach lets you safely close the dialog if you want to stash your changes manually. I'm on the side of love-linger here. 
remembering options for these actions is asking for trouble. Sane defaults are the better way to go combined with actual preference choices that are remembered (rebase vs merge on pull for example). But anything that is potentially dangerous or never a 'sane default' should not be remembered

sven-n:
I've had a similar problem several times this week. I was in a hurry and didn't pay too much attention to what option was selected for local changes when doing a pull or merge. Unfortunately it was 'discard', which sourcegit remembered from last time. Can we at least agree that it's a bad idea to pre-select 'discard'?

mithom:
> Can we at least agree that it's a bad idea to pre-select 'discard'?
that, i can agree with, although i'm very happy that the stash & apply option is remembered (my company uses rebasing a lot which makes this a required feature to happily work with git)

love-linger:
I decide not to record the last selected handling method for uncommitted local change anymore. Instead, the default option is `Do Nothing` (let git decide how to handle it - when there are modified files, git will prevent related operations). If necessary, users can change the handling method according to the actual situation on their own.

mithom:
[...]

PR #1516, opened 2025-07-04
#977 - Discarded changes should be saved in a dated file in the Trash

xavierdono:
Tested on Win 11 only.
Need to be test on Linux and MacOS.

FYI: It would be more easier to put the file in a folder (choose by the user, need to change the pref) that in the trash due to multiplatform.


xavierdono:
I better understand the codebase, I use buit it solution (OS detection and Move file to trash interface).

PR #1517, opened 2025-07-04
enhance: support tools installed via flatpak

NathanBaulch:
This PR is a bit experimental.
I have various tools installed via [Flatpak](https://flatpak.org) (Meld, VSCodium, etc) that I'd like sourcegit to integrate with. Currently all diff and merge tools must resolve to a known binary file, whereas Flatpak apps are launched via the flatpak binary (eg `flatpak run org.gnome.meld`). I've figured out some workarounds to get `ExternalTool` and `ExternalMerger` working but I'm reluctant to make any significant changes because I can only test on Linux.

### DIFF
--- a/src/Native/Linux.cs
+++ b/src/Native/Linux.cs
@@ -13,6 +13,8 @@ namespace SourceGit.Native
     [SupportedOSPlatform("linux")]
     internal class Linux : OS.IBackend
     {
+        private static readonly string LOCAL_APP_DATA_DIR = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+
         public void SetupApp(AppBuilder builder)
         {
             builder.With(new X11PlatformOptions() { EnableIme = true });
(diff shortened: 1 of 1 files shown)


love-linger:
I suggest directly creating a command-line tool for `Flatpak` applications, so it can be used as normal application. For example:

1. Create `/usr/bin/meld`，and fill it with following contents：
  ```
  #!/bin/bash
  
  /usr/bin/flatpak run org.gnome.meld $@
  ```
2. Change the file mode with `chmod u+x /usr/bin/meld`

NathanBaulch:
Digging deeper, it looks like Flatpak already does this for me! Specifying the following install path works for my per-user Meld installation: `/home/nathan/.local/share/flatpak/exports/bin/org.gnome.meld` 👍
It would be cool if sourcegit could figure this out automatically, but at least I have something workable.
It would also be cool if it could automatically detect these in `ExternalTool`, which is not something I can override in preferences.

NathanBaulch:
I've updated this PR with much cleaner Flatpak external tool support based on inspecting the system-wide and per-user `flatpak/exports/bin` folders. Changes are limited to `Linux.cs` and `ExternalMerger` is no longer touched.

If sourcegit can have special support for JetBrains Toolbox then it seems reasonable that it can also support Flatpak on Linux.

love-linger:
> If sourcegit can have special support for JetBrains Toolbox then it seems reasonable that it can also support Flatpak on Linux.

It's because this is a cross-platform app, and there are so many JetBrains products (currently we support 16 JetBrains products), and their installation paths vary on different platforms. Also, different versions can coexist. To solve the startup problem, JetBrains itself provides the JetBrains Toolbox. This is not exactly the same as third-party package managers like Flatpak.

NathanBaulch:
I agree with everything you've said.
I still don't see any reason why the Linux backend can't search in a couple of extra places to find tools installed with a popular distro-agnostic package manager. The same goes for tools installed via Mac App Store or Microsoft Store if they can be handled by their respective `IBackend`.

Issue #822, opened 2024-12-16
Feature request: Display current branch and number of uncommitted files in repositories tab


kethinov:
Kinda related to https://github.com/sourcegit-scm/sourcegit/issues/821

It would also be nice to see the current selected branch and how many files are not yet committed, e.g. see the below screenshot from Sourcetree.

![Screenshot 2024-12-16 at 5 14 54 PM](https://github.com/user-attachments/assets/b96ace6b-0114-4115-abd0-57fd776624d3)

PR #1528, opened 2025-07-05
#822 - Display current branch and number of uncommitted files in repositories tab

xavierdono:
I use the command : "git log origin/develop..HEAD --name-only --pretty=format:", hope is good.


love-linger:
Uncommitted local changes are already displayed in `LOCAL CHANGES` page switcher.

PR #1535, opened 2025-07-06
localization: move strings into resources

NathanBaulch:
Found 22 static strings that weren't localizable. I took a guess at the key names, feel free to rename.

### DIFF
--- a/src/Views/InteractiveRebase.axaml.cs
+++ b/src/Views/InteractiveRebase.axaml.cs
@@ -161,17 +161,17 @@ namespace SourceGit.Views
             flyout.Placement = PlacementMode.BottomEdgeAlignedLeft;
             flyout.VerticalOffset = -4;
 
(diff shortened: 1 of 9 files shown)


xavierdono:
App.Text add "Text." before.

App.Text("Text.InteractiveRebase.Pick") it's just App.Text("InteractiveRebase.Pick"), no?

NathanBaulch:
Good point @xavierdono, fixed!

src/Views/InteractiveRebase.axaml.cs
```diff
@@ -140,17 +140,17 @@ private void OnButtonActionClicked(object sender, RoutedEventArgs e)
             flyout.Placement = PlacementMode.BottomEdgeAlignedLeft;
             flyout.VerticalOffset = -4;
 
-            CreateActionMenuItem(flyout, Brushes.Green, "Pick", "Use this commit", "P", item, Models.InteractiveRebaseAction.Pick);
```
love-linger, line 143: Action name like `Pick`, `Squash` etc. should not be translated.

PR #1541, opened 2025-07-07
Add Blame command to WorkingCopyContextMenu

oikku:


love-linger:
For `File/Dir History`, it is understandable to enable it in `LOCAL CHANGES` because they show all the commits in history that changes the selected file/dir exclude the current state of file (since it has not been committed). However, for `Blame`, it is not appropriate to place it here. The reasons are:

- Blame shows who last modified each line of a file, along with the commit hash and timestamp for each change. But it can not include the changes in current worktree.
- It will not work for untracked files, but it can work for deleted file in current worktree (not commited yet)

oikku:
With large project having 20 years of history and multiple editors I find myself using blame when editing file. You start making changes and at some point you encounter code you really don't understand why it is done how it is done and at that point you want to see the blame of original file before your changes to get better understanding of the situation. Now when blame is missing from local changes extra work has to be done to open that blame view.

Untracked files could be skipped just like directories and it could be shown to deleted files.

PR #1543, opened 2025-07-07
code_style: general cleanup

NathanBaulch:
Another batch of cleanup.


NathanBaulch:
Feedback addressed in a new commit so you can see the followup fixes in isolation.

PR #1544, opened 2025-07-07
enhance: rebase move up/down with arrow keys

NathanBaulch:
I find reordering commits with the mouse a bit finicky, especially since you have to stay within the drag/drop anchor column. Much easier to move commits with the keyboard.


love-linger:
Arrow up/down is used to select prev/next commit in list. It's more common to users than re-order commits.

NathanBaulch:
Ctrl+Up/Down maybe?

love-linger:
> Ctrl+Up/Down maybe?

I don't think there is any problem with this.

NathanBaulch:
Great, done. There's nothing in the UX to inform the user of this ability however.

PR #1551, opened 2025-07-09
enhance: tweak `ToFileSize` for readability

NathanBaulch:
The new file size format has too many significant digits for good readability, plus thousands separators would help for the raw byte count.
Also, is now a good time to start adding some unit tests?

```
Before: 15.103 KB (15465 B)
After:  15.1 KB (15,465 B)

Before: 491.268 KB (503058 B)
After:  491 KB (503,058 B)

Before: 1.465 MB (1536332 B)
After:  1.46 MB (1,536,332 B)
```

### DIFF
--- a/.github/workflows/tests-check.yml
+++ b/.github/workflows/tests-check.yml
@@ -1,0 +1,24 @@
+name: Format Check
+on:
+  push:
+    branches: [develop]
+  pull_request:
+    branches: [develop]
+  workflow_dispatch:
+  workflow_call:
+
+jobs:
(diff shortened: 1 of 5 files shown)


love-linger:
I've pushed my implementation for this request which based on current culture

NathanBaulch:
Your implementation doesn't address this PR - too many digits in the formatted file size.
For example it now shows `491.268 KB` when `491 KB` is all the user needs to see. `G3` is much better than `N3` in this situation.

Issue #1519, opened 2025-07-04
[Feature] add visual studio preview to external tools


FreyLuis:
I'd love to be able to use both Visual Studio and Visual Studio Preview as external tool options.

Since the registry keys that Visual Studio creates seem... messy, I don't think the current approach will work reliably.

Microsoft made a small utility called vswhere, which is specifically designed to locate installed Visual Studio instances.
The exe is available via. [nuget](https://www.nuget.org/packages/vswhere) which we could use to include it in the output for windows builds.

If that sounds like a good approach, I'd be happy to make a PR.

Output Example:
```
> .\vswhere.exe -prerelease
Visual Studio Locator version 3.1.7+f39851e70f [query version 3.12.2140.44225]
Copyright (C) Microsoft Corporation. All rights reserved.

installationName: VisualStudioPreview/17.14.7-pre.1.0
installationPath: C:\Program Files\Microsoft Visual Studio\2022\Preview
productPath: C:\Program Files\Microsoft Visual Studio\2022\Preview\Common7\IDE\devenv.exe
isPrerelease: 1

installationName: VisualStudio/17.14.7
installationPath: C:\Program Files\Microsoft Visual Studio\2022\Professional
productPath: C:\Program Files\Microsoft Visual Studio\2022\Professional\Common7\IDE\devenv.exe
isPrerelease: 0
```

PR #1553, opened 2025-07-10
feat: detect visual studio preview as external tools

FreyLuis:
This PR adds detection for Visual Studio Preview, as described in #1519.

### DIFF
--- a/src/Native/Windows.cs
+++ b/src/Native/Windows.cs
@@ -2,10 +2,11 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
 using System.Text;
-
+using System.Text.RegularExpressions;
 using Avalonia;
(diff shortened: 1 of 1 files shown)


FreyLuis:
The vswhere binary is installed with the Visual Studio Installer.

Alternatively, we could ship with the vswhere binary with the following:
```xml
  <!-- Testing for windows as target platform. fallback to true when no identifier is set so a nuget restore still works -->
  <ItemGroup Condition="'$(RuntimeIdentifier)' == '' OR $([MSBuild]::StartsWith('$(RuntimeIdentifier)', 'win'))">
    <PackageReference Include="vswhere" Version="3.1.7" GeneratePathProperty="True" PrivateAssets="all" />
    <Content Include="$(PkgVswhere)\tools\vswhere.exe" Visible="False" Link="vswhere.exe" CopyToOutputDirectory="PreserveNewest" />
  </ItemGroup>
```

But this seems to work fine.

love-linger:
I've pushed my implementation of this PR that based on your idea.

PR #1556, opened 2025-07-11
enhance: support `--autostash` on interactive rebase

NathanBaulch:
I often have minor patches in my workspace that I'd like to keep during interactive rebase. This PR adds the same autostash ability already available in non-interactive rebase.

It would also be nice to allow commit rename with pending changes so long as nothing is staged, however I couldn't find an easy way to determine this from the `Histories` view model.

### DIFF
--- a/src/Commands/Rebase.cs
+++ b/src/Commands/Rebase.cs
@@ -15,12 +15,15 @@
 
     public class InteractiveRebase : Command
     {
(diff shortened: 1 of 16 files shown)


love-linger:
If you mean to reword the HEAD commit, now you can do it while keeping local changes. You can also do it by just commit with `--amend` enabled.
image

Issue #1424, opened 2025-06-16
[Feature] Support .issuetracker file


andgeno:
It would be great if SourceGit would support automatic issue tracker setup by checking for a `.issuetracker` file in the repo root dir.

Example of a `.issuetracker` file:
```ini
# Integration with Issue Tracker
#
# (note that '\' need to be escaped).

[issuetracker "Azure DevOps"]
  regex = "#(\\d+)"
  url = "https://dev.azure.com/<yourname>/_workitems/edit/$1"
```

PR #1567, opened 2025-07-13
feature: support `.issuetracker` file

sina-hide:
Beforehand issue trackers could be defined in the repository configuration only.  They were stored in the `.git` directory.  As a consequence each SourceGit user had to configure issue trackers for himself.

Now additional issue tracker rules are read from an existing `.issuetracker` file (#1424).  An issue tracker configured by a user has a higher priority than an issue tracker read from `.issuetracker`.

This allows to share issue trackers between team members.

SourceGit does not (yet?) write into the `.issuetracker` file.  But using `git config` allows the user to easily construct issue tracker rules for the `.issuetracker` file, where even the escaping of the backslash in '#(\d+)' is done correctly.

A `.issuetracker` file was added to SourceGit.  It was created by executing the following commands in a bash shell, where the current directory was the repository root:

```bash
cat >.issuetracker <<EOF
# Integration with Issue Tracker
#
# (note that '\' need to be escaped).

EOF
git config set --file .issuetracker issuetracker.GitHub.regex '#(\d+)'
git config set --file .issuetracker issuetracker.GitHub.url \
  'https://github.com/sourcegit-scm/sourcegit/issues/$1'
```

### DIFF
--- a/.issuetracker
+++ b/.issuetracker
@@ -1,0 +1,7 @@
+# Integration with Issue Tracker
+#
+# (note that '\' need to be escaped).
+
+[issuetracker "GitHub"]
+	regex = "#(\\d+)"
+	url = https://github.com/sourcegit-scm/sourcegit/issues/$1
--- a/src/Commands/Config.cs
+++ b/src/Commands/Config.cs
@@ -20,9 +20,9 @@ namespace SourceGit.Commands
             }
         }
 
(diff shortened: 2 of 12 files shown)


sina-hide:
I've rebased onto the current `develop` branch.

love-linger:

Now you can share issuetracker rules by enable `Share this rule in .issuetracker file` option

image

PR #1570, opened 2025-07-14
enhance: more readable command log duration

NathanBaulch:
Minor tweak to the command log duration format, another good use for `G3`!


love-linger:
If the time exceeds 1000s (such as cloning a large repo), `G3` will use scientific notation to represent it.

image

NathanBaulch:
I assumed 99% of operations would be under a minute, and nothing would exceed 999 seconds. Let's just add minute formatting too.

PR #1572, opened 2025-07-14
enhance: quickly drop single commits by prefilling interactive rebase

NathanBaulch:
I often find myself wanting to drop (not revert) a single commit. It would be nice if this could be streamlined - by launching the existing interactive rebase window with the relevant commit already marked as `Drop`.


love-linger:
I've pushed my implementation for this feature request.

NathanBaulch:
I was hoping you'd come up with a more complete solution @love-linger, thanks!

PR #1573, opened 2025-07-14
localization: update spanish translations

jjesus-dev:
Add missing translations.

This lines were previously translated but still appear in TRANSLATION.md as missing. @love-linger 
- Text.CommitCM.InteractiveRebase
- Text.CommitCM.Rebase
- Text.CommitCM.Reset
- Text.CommitCM.SquashCommitsSinceThis


love-linger:
It's because PR #1568 has changed the original meaning of these lines

jjesus-dev:
> It's because PR #1568 has changed the original meaning of these lines

Oh, ok. I will check later, thanks.

sina-hide:
Since #1461 @love-linger removes translations that changed its meaning.  So it's obvious which translations have to be updated.

PR #1589, opened 2025-07-16
feat: Add UserActivityTracker to monitor user activity

massimopaganigh:
Introduce a new `UserActivityTracker` class in the `SourceGit.Models` namespace to track user activity and determine when to perform auto-fetch operations. Update the `App.axaml.cs` file to initialize the tracker at startup and modify `Repository.cs` to utilize this new functionality, ensuring auto-fetching occurs only during user inactivity.


massimopaganigh:
The aim of pr is to prevent user-defined operations from overlapping with automatic operations causing error notifications.

PR #1590, opened 2025-07-17
fix: Support commit template with relative path

ooooo84:


love-linger:
This PR contains so many unexpected changes.

- Do not using namespace in `SourceGit.*` since there's so many classes use the same name in different namespace, such as `SourceGit.Models.Branch`, `SourceGit.Commands.Branch`
- Unnecessary newlines (Code style changes)

PR #1592, opened 2025-07-17
feat: add color element to local changes count to increase visibility

scotc-dev:
I always felt that the local changes does not have enough visual feedback to to show changes

### DIFF
--- a/src/Converters/IntConverters.cs
+++ b/src/Converters/IntConverters.cs
@@ -39,5 +39,21 @@ namespace SourceGit.Converters
                 else
                     return Models.Bookmarks.Brushes[bookmark];
             });
+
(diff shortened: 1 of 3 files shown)


scotc-dev:
image

goran-w:
NOTE: The blueish highlight is elsewhere used to indicate a selected tab/item (like the Home icon in your screenshot) or an activated option (for toggle-buttons), so IMHO it looks a bit misleading when used on a count-badge that is just there for indication.

scotc-dev:
It might just be me, I just find the white on grey to be hard to see.

goran-w:
> It might just be me, I just find the white on grey to be hard to see.

If that is a general concern / preference, you might want to try a custom color-theme (or create your own) with more contrast :
https://github.com/sourcegit-scm/sourcegit-theme

PR #1608, opened 2025-07-21
enhance: Add "Move to New Window" and "Copy to New Window" options

massimopaganigh:
Enhanced the user interface by adding new menu items for "Move to New Window" and "Copy to New Window" in various language files. Implemented the corresponding functionality in `Launcher.cs` to allow users to move or copy tabs to new windows.

image


love-linger:
I don't plan to add support for multi-instance or multi-window features now since there are lots of bugs in this PR:

- The `App` only holds a single instance of `ViewModels.Launcher`, which is used to dispatch notification or exceptions
- `ViewModels.Repository.GetOwnerPage` does not work in this PR
- In this PR, if the user cannot determine which is the first window, it is easy to close the first one and cause the program to exit.
- The `Workspace` feature does not work correctly in this mode.

Issue #1632, opened 2025-07-29
[FR]: 编辑动作提供复制选项/add Duplicate CustomAction


heartacker:
<img width="615" height="566" alt="Image" src="https://github.com/user-attachments/assets/d6533446-dab0-4b6f-9ff8-b3caa68b47ef" />

PR #1634, opened 2025-07-29
add DuplicateCustomAction( #1632)

heartacker:


heartacker:
PixPin_2025-07-30_11-20-28

goran-w:
NOTE: A similar request (for a way to Clone a Custom Action) was recently brought up in this Discussion: #1479

goran-w:
@love-linger, @heartacker - This is an excellent (and very simple / straight-forward) feature which deserves merging ASAP. 

However, it's so great that we should have the same feature for `Issue Tracker` rules (and `Commit Template` items) as well! 😉

heartacker:
@love-linger 已经根据社区意见修改。麻烦看看这个是否有新的看到，并review 下。谢谢哈

love-linger:
我并不想加入该功能。我认为在一个仓库中复制一个自定义动作是无意义的（它们不可能相同，且重新编辑一个又不耗费时间）。

I don't want to add this feature. I think it's meaningless to copy a custom action in a repository (they can't be the same, and it doesn't take time to re-edit one).

love-linger:
对于此类需求，我自己来实现的话会是增加一个拷贝的操作（复制到剪贴板），允许从剪贴板创建。这样，就可以将某一个项目中已配置好的某些自定义操作直接拷贝到另一个仓库。

For such requirements, if I were to implement it myself, I would add a copy operation (copy to the clipboard) and allow creating a new one from the clipboard. In this way, user can copy existing custom action from one repository to another.

heartacker:
其实主要是有参数配置的部分会比较麻烦，复制来复制去，挺不自在的

尤其是有好几个参数，有描述有定制，一个个复制很麻烦。甚至很记住


当然，如果是一行命令行，那确实很简单。

heartacker:
> 增加一个拷贝的操作（复制到剪贴板），允许从剪贴板创建。

也是很好的主意，期待新的PR,

如果是这样的话，一个复制的按钮，也是不错的的，可以在用户点击复制按钮的时候，选复制到本repo 还是复制到 剪切板。

goran-w:
> For this type of requirement, my own implementation would involve adding a copy operation (copy to clipboard) and allowing creation from the clipboard. This way, certain custom operations configured in one project can be directly copied to another repository.

I would much prefer having a Save & Load mechanism for transfer of `Custom Actions` / `Issue Tracker` rules / `Commit Templates` between repos, since that would also work between machines!

I would still, however, also keep the "Clone" mechanism proposed in this PR, since it's not for transfer but for quicker experimenting and creation of several similar or related actions / rules / templates.

goran-w:
> I don't want to add this feature. I think it's meaningless to copy a custom action in a repository (they can't be the same, and it doesn't take time to re-edit one).

It's not meaningless, since the purpose is to make small changes to an existing Action in order to make a new variant.

Also, it can certainly take time to "re-edit" one, since it can potentially have several Input Controls and multiple command arguments (with variables) etc.

(For example, I recently created two very similar Actions, with nearly identical inputs and args but with different scope : "Rename remote branch" and "Rename remote tag". Having a Clone mechanism would have been very helpful when creating the 2nd one, since it was a bit tedious to copy individual settings from the 1st one.)

That's why I suggest two complementary mechanisms:
* A "trivial" Clone mechanism (as in this PR) for quick-and-easy copy-then-edit functionality within the same editor panel.
* A new Save & Load feature, for transfer of Actions between repo and global scope, between repo-tabs and between computers/users.

For consistency, the same type of functionality (at least the Clone mechanism) should be added to the editors for Issue Tracker rules and for Commit Templates.

heartacker:
是否可以将这个合并，我长时间使用后，我觉得这个是一个非常实用的功能。然后和配置分享的功能也不冲突，我们可以在后续继续开发 配置分享的功能。
---
Can this be merged? After long-term use, I find this to be a highly practical feature. Additionally, it does not conflict with the configuration sharing function, and we can continue to develop the configuration sharing feature in subsequent iterations.

PR #1642, opened 2025-07-31
feat: support launch a debug sourcegit

heartacker:


love-linger:
As I mentioned in the several ISSUEs about multi-instance of `SourceGit`, running multiple instances currently leads to some unpredictable issues, such as the saving problem of `preferences.json`. 

Therefore, whether in Debug or Release mode, `SourceGit` currently only allows one instance to run.

heartacker:
ok

PR #1648, opened 2025-08-01
feat: remember fetch all remotes property(fixes #1647)

### LINKED ISSUE #1647: [FR] 记住 fetch 时的`拉取所有的远程仓库`的选项
每次都是默认check，每次都得点击取消他，

Image

因为有多个 remote ，有的remote 需要密码，我仅仅是用来备份用的，不需要 fetch， 不然每次都要我输入密码太烦了，所以我不需要fetch他
请不要每次都check 这个按钮

heartacker:
- Add a new property to enable fetching all remotes by default.
- Refactor fetch logic to use repository settings directly instead of local properties.

### DIFF
--- a/src/Models/RepositorySettings.cs
+++ b/src/Models/RepositorySettings.cs
@@ -62,6 +62,12 @@ namespace SourceGit.Models
             set;
         } = false;
 
+        public bool FetchAllRemotes
+        {
+            get;
+            set;
+        } = false;
+
         public bool FetchWithoutTags
         {
             get;
--- a/src/ViewModels/Fetch.cs
+++ b/src/ViewModels/Fetch.cs
@@ -18,9 +18,13 @@ namespace SourceGit.ViewModels
 
         public bool FetchAllRemotes
         {
-            get;
-            set;
-        } = false;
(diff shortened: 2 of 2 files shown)


love-linger:
If you want to just fetch specific remote, you should use context menu entry of selected remote:
image

heartacker:
If we're simply discussing the  git fetch  function, along with its corresponding button and shortcut key, I think what we should focus on is its pure original meaning. Its behavior should be as simple as just entering  git fetch  in the terminal. Frankly speaking, I don't think  all  should be set as the default option.
 
Although we can start from the  remote  side and only  fetch  a specific  remote , it's really not that fast. I just want  git fetch  and its original behavior, and have a shortcut key for it.

heartacker:
The  fetch  button and its shortcut key should follow its original behavior. It should  fetch  the default remote. And that's exactly what we want. Why does it have to be  fetch all  and set as the default permanently? It should rather cover the basic and original meaning. Moreover, after using  fetch all , there may be inconveniences in some scenarios, such as the need for password verification.
 
If there is an option, I think  all  should be disabled，by default

PR #1653, opened 2025-08-02
feat: add Built-in parameters: `DIRNAME`,  #1628, #1633

heartacker:


love-linger:
You should use working directory of the custom action's process, which is always the path of repository

heartacker:
we just want the `folder` name, 
the `folder` name is fix and part of `path`, 
dirname remains unchanged, just like path, and we also need to use dirname for better processing
please

heartacker:
I know the path is fixed, but I only need the name of the folder. And this folder is fixed just like the path, which is very useful for custom processing. I need to get the name of the folder instead of the name of the path

love-linger:
```shell
DIR_NAME=$(basename `pwd`)
```

heartacker:
it can not work in action command, 

```
DIR_NAME=$(basename `pwd`)
```

PixPin_2025-08-04_10-52-40

heartacker:
https://github.com/sourcegit-scm/sourcegit/issues/1629

heartacker:
PixPin_2025-08-04_11-20-29

love-linger:
It's a shell expr. You should use ```echo "$(basename `pwd`)"```

love-linger:
image

image

heartacker:
That is to say, even for further customized requirements, even if they are simple, I have to use the shell. Moreover, under different operating systems, such as Linux or Windows, I also have to write specific and different commands for the operating systems to achieve the same function.

Saibamen (approved):

PR #1660, opened 2025-08-04
fix: exclude main worktree by comparing with commonDir in submodules

gadfly3173:
Fixes worktree detection in submodules where worktree.FullPath differs from commonDir path.


love-linger:
I didn't understand what problem this PR is targeting. Could you provide a comparison of the output results of a demo?

gadfly3173:
for a submodule :
real path: `src/__theme__`
common dir (worktree output): `.git/modules/src/__theme__`

gadfly3173:
image

PR #1672, opened 2025-08-06
respect `$BROWSER` env in linux

heartacker:


gadfly3173:
@love-linger I think this is useful and can be merged.

PR #1724, opened 2025-08-12
Make NumericSort use InvariantCulture for text comparison

goran-w:
Changed the comparison method used in `NumericSort.Compare()` :
* `Ordinal` comparison sorts non-alphanumeric chars (like underscore etc) differently than expected (i.e compared to filesystem sorting etc). 
* `InvariantCulture` comparison corrects this, without imposing cultural differences.
* (A possible alternative would be `CurrentCulture` comparison.)

Additionally:
* When comparing a single digit-char against a single non-digit-char, the earlier code used `char.CompareTo()` where the comparison method can not be specified.
* To fix this, we now use the same consistent `Comparer` object for all the string comparisons which are NOT "digits vs digits".

### DIFF
--- a/src/Models/NumericSort.cs
+++ b/src/Models/NumericSort.cs
@@ -6,6 +6,8 @@ namespace SourceGit.Models
     {
         public static int Compare(string s1, string s2)
         {
+            var comparer = StringComparer.InvariantCultureIgnoreCase;
+
             int len1 = s1.Length;
             int len2 = s2.Length;
 
@@ -20,7 +22,7 @@ namespace SourceGit.Models
                 bool isDigit1 = char.IsDigit(c1);
                 bool isDigit2 = char.IsDigit(c2);
                 if (isDigit1 != isDigit2)
-                    return c1.CompareTo(c2);
+                    return comparer.Compare(c1.ToString(), c2.ToString());
 
                 int subLen1 = 1;
                 while (marker1 + subLen1 < len1 && char.IsDigit(s1[marker1 + subLen1]) == isDigit1)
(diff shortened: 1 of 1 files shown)


goran-w:
@love-linger NOTE: This PR does NOT change the behavior of numeric comparisons in the NumericSort class.

My change only improves how non-numeric comparisons are made (to handle symbols like underscore in a way that's more consistent with the filesystem etc).

Also, it fixes an actual bug where the single-char (digit vs non-digit) case used a different comparison method than the normal (non-numeric) substring case!

Finally, by using a comparator object for the non-numeric comparisons, it becomes very easy to change the comparison-method used (if needed).

love-linger:
https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings

image

I want to write a relevant performance test result, but I don't have time recently.

goran-w:
An alternative could be to use `CurrentCulture` : "when you display output to the user", as mentioned in the documentation section you quoted... (I was trying to use something that was closer to `Ordinal`, i.e culture-agnostic, but still did something more "intelligent" than sorting chars by their underlying char-code like `Ordinal` seems to do.)

For example, when sorting with `OrdinalIgnoreCase`, names beginning with underscore are sorted at the bottom rather than at the top like in the filesystem (Windows Explorer etc). Using either `InvariantCultureIgnoreCase` or `CurrentCultureIgnoreCase` fixes this.

Example of sort-order with `OrdinalIgnoreCase` : 
image

Example of sort-order with `InvariantCultureIgnoreCase` or `CurrentCultureIgnoreCase` (and in filesystem listings etc) :
image

BTW, I don't think this has a noticeable effect on performance - doing the "Numeric Sorting" (at all) probably has a larger impact anyway?

PR #1728, opened 2025-08-13
Use env variable for update detection

flipper:
This allows update detection to be disabled using a env var instead of a compilation define.

### DIFF
--- a/src/App.Commands.cs
+++ b/src/App.Commands.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Windows.Input;
 using Avalonia.Controls;
+using SourceGit.Native;
 
 namespace SourceGit
 {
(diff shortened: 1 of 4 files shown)


love-linger:
This PR will not be merged. It is because:

* User can simply disable the update detection feature from `Preferences` window if they download the `SourceGit` from GitHub Releases page of this repository.
* We need a way to remove the update detection feature completely if someone want to publish their own build to a package manager. For example: [SourceGit@AUR](https://aur.archlinux.org/packages/sourcegit)

flipper:
> This PR will not be merged. It is because:
> 
>     * User can simply disable the update detection feature from `Preferences` window if they download the `SourceGit` from GitHub Releases page of this repository.
> 
>     * We need a way to remove the update detection feature completely if someone want to publish their own build to a package manager. For example: [SourceGit@AUR](https://aur.archlinux.org/packages/sourcegit)

Yes that is correct. Will close the PR!

PR #1741, opened 2025-08-18
fix: migrate issue trackers of 2025.26

### LINKED ISSUE #1725: [Bug] Jira and GitHub Issue Tracker settings keep getting deleted every time SourceGit 2025.29 is launched
SourceGit is a good app. However, whenever SourceGit 2025.29 is launched, it will delete all of the settings for the Jira Issue Tracker links or GitHub Issue Tracker links. This is bad because it results in data loss, and the user becomes frustrated when they have to spend a significant amount of time adding the settings back.

### Steps to reproduce the bug, using SourceGit 2025.29 App Image

1. Launch `sourcegit-2025.29.linux.amd64.AppImage` on Ubuntu Linux. (Other systems may be affected. I have not tested them)
2. Open a Git repository that contains some history. A good example is the SourceGit repository on GitHub.
3. Click on the button `"Configure this Repository"` at the top-left

Image

4. Go to the "Issue Tracker" tab and add some settings for either GitHub or Jira.

Image

5. Close and relaunch SourceGit 2025.29 and look at the `"Configure this Repository" -> "Issue Tracker"` settings again

### Expected Result

* All the settings should be preserved from what I entered the last time.
* No settings should be deleted.

### Actual Result

* The settings are completely erased for every repository.

Image

* The user has to spend a large amount of time re-adding these settings for each Git repo. But they will just delete themselves anyway when the app is Quit and reopened.
* These settings are not saved in the 'Data Storage Directory' in `/home/username/.config/SourceGit/preference.json`

### Impact on customers

* It causes a large data loss due to forgotten settings.
* Users get annoyed because it takes a long time to re-add these settings, even if copying them from a text file.
* Older versions of SourceGit did not have this problem. Examples are `sourcegit-2025.14.linux.amd64.AppImage`
* The same problem also happens when upgrading SourceGit from an older version to 2025.29. All the Issue Tracker settings are lost.

sina-hide:
Starting with version 2025.26 (implemented in commits b345bb4 and 9802ac9 for issue #1599) the issue trackers configured in repository settings are stored in a new format, deleting the issue trackers configured before.

Since this breaks compatibility, now these issue trackers in the old format used in versions up to 2025.26 are automatically migrated, so they don't get lost.

Fixes #1725.

### DIFF
--- a/src/App.JsonCodeGen.cs
+++ b/src/App.JsonCodeGen.cs
@@ -65,6 +65,7 @@ namespace SourceGit
     [JsonSerializable(typeof(Models.ThemeOverrides))]
     [JsonSerializable(typeof(Models.Version))]
     [JsonSerializable(typeof(Models.RepositorySettings))]
+    [JsonSerializable(typeof(Models.IssueTrackerMigrator.IssueTrackerRepositorySettingsVersion202526))]
     [JsonSerializable(typeof(List<Models.VisualStudioInstance>))]
     [JsonSerializable(typeof(ViewModels.Preferences))]
     internal partial class JsonCodeGen : JsonSerializerContext { }
--- a/src/Models/IssueTrackerMigrator.cs
+++ b/src/Models/IssueTrackerMigrator.cs
@@ -1,0 +1,58 @@
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+using Avalonia.Collections;
+
+namespace SourceGit.Models
+{
+    public static class IssueTrackerMigrator
+    {
+        public static IEnumerable<IssueTracker> Migrate(string gitDir)
+        {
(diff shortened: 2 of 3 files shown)


sina-hide:
@love-linger This code is written in a way that it should be easy to revert if the code no longer deems necessary.  Maybe in a year, when most people should use versions after 2025.30.

love-linger:
I'm sorry. This PR will not be merged.

* The storage location of the local IssueTracker has changed several times. Before `2025.28`, it was stored in `$GIT_DIR/sourcegit.settings`, and on `2025.28`, it was changed to `$GIT_DIR/sourcegit.issuetracker`, and now it has been moved to the default `$GIT_DIR/config`. This modification has no effect on versions that have been updated to `2025.28` and later.
* The `Issue Trackers` configuration is not a very important and difficult-to-restore configuration.
* I will emphasize this issue in the version description.

PR #1771, opened 2025-08-27

jjesus-dev:
- Add missing strings

Hi @love-linger , I need more context about the `Text.Preferences.AI.ReadApiKeyFromEnv` string. If I understand correctly, there is an API key located in ENV and you need to input that variable name to load it's value?

If this is correct, I going to adapt the spanish version a little bit because the literal translation is confusing with the phrasing in spanish.


love-linger:
> I need more context about the Text.Preferences.AI.ReadApiKeyFromEnv string. If I understand correctly, there is an API key located in ENV and you need to input that variable name to load it's value?

Yes. You are right.

PR #1807, opened 2025-09-10
feat: Visual Studio 2026 icon support

massimopaganigh:


love-linger:
The icon here represents a type of tool and is unrelated to the specific version information of the tool (VS/VSCode will distinguish whether it is a insider/preview version).

Issue #1605, opened 2025-07-19
Feature request: Add option to set depth for fetch / pull


silmaril42:
When working with very large repos (eg. https://github.com/NixOS/nixpkgs), fetching or pulling everything may take very long and get huge amounts of data that isn't needed at all.

While this can already be addressed by adding `--depth 10` or similar as "Extra Parameters" in the clone dialogue, I'd like to have such an option for fetch and pull, too.

The syntax for fetch and pull is slightly different from clone: we need an `=` instead of the space: `--depth=10`.

I could think of several ways to implement this:

1. Add a `depth` field to the fetch/pull dialogues (default: empty or `0`, meaning "fetch everything"). This setting should be remembered.
2. Add an `Extra Parameters` field to the fetch/pull dialogues, similar to the clone dialogue.
3. Add a field `Fetch / Pull depth` (or two separate fields, if that makes any sense) to the Repository Configure dialogue and use this parameter for all fetch / pull operations.

I'm not sure, which variant is the best idea. Personally, I like option 3 best, but all of them would be fine for me.

massimopaganigh:
Considering the possibility of having the auto fetch enabled, I implemented option 3 with separate parameters for fetch and pull.

love-linger:
I recommanded you to create a CUSTOM ACTION to do this

PR #1817, opened 2025-09-20
[feature] Repository depth ( --depth argument )

Cappecasper03:
Implemented depth setting that can be saved globally or per repository and chosen for each fetch, pull and clone

### DIFF
--- a/src/Commands/Clone.cs
+++ b/src/Commands/Clone.cs
@@ -10,7 +10,7 @@
             Args = "clone --progress --verbose ";
 
             if(depth > 0)
-                Args += $"--depth {depth} ";
+                Args += $"--depth {depth} --no-single-branch ";
 
             if (!string.IsNullOrEmpty(extraArgs))
                 Args += $"{extraArgs} ";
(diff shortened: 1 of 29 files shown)


Cappecasper03:
@Mingun Is this how it should be?

massimopaganigh:
https://github.com/sourcegit-scm/sourcegit/pull/1609
https://github.com/sourcegit-scm/sourcegit/issues/1605

Cappecasper03:
So this is not planned and I should just use a custom action instead?
The part with custom actions is that if I accidentally use the regular fetch/pull, I would get everything so it would be nice to just have a setting that all actions use to avoid any accidents

massimopaganigh:
> So this is not planned and I should just use a custom action instead?
> The part with custom actions is that if I accidentally use the regular fetch/pull, I would get everything so it would be nice to just have a setting that all actions use to avoid any accidents

I know, but I had already tried to make a mr in the past, unfortunately since the related issue was closed and since there was no activity on the mr I decided to close it.

PR #1827, opened 2025-09-24
added copy button next to the commit message inside the information tab

evilbaschdi:


love-linger:
image

evilbaschdi:
Oh, I did not know 'bout that possibility.

I already knew that I can copy the message right from the "Graph" but for me that was "one more click".
So I've added that button. That context menu you're showing is also one more click but provides more options. So I'll close that PR.

BWT: Is there a documentation listing all functionaliies? Since I've switched to SourceGit from another tool recently which had the feature of untracking files the next thing I would have worked on is untracking files (and maybe also folders) from that "FILES" tab next to INFORMATION and CHANGES. So I'm just asking before that would be the next PR to be closed :-D

PR #1831, opened 2025-09-27
Translation Utility Script

ramonmelo:
## Translate Helper

> A script to help with translations by reading the target language, comparing with the base language, and going through missing keys.

### Usage

```bash
python translate_helper.py pt_BR [--check]
```

- `pt_BR` is the target language code (change as needed)
- `--check` is an optional flag to only check for missing keys without prompting for translations

The script will read the base language file (`en_US.axaml`) and the target language file (e.g., `pt_BR.axaml`), identify missing keys, and prompt you to provide translations for those keys. If the `--check` flag is used, it will only list the missing keys without prompting for translations.

### DIFF
--- a/utils/README.md
+++ b/utils/README.md
@@ -1,0 +1,18 @@
+# Utility scripts
+
+> collection of utility scripts for various tasks
+
+## Translate Helper
+
+> A script to help with translations by reading the target language, comparing with the base language, and going through missing keys.
+
+### Usage
+
+```bash
+python translate_helper.py pt_BR [--check]
+```
+
+- `pt_BR` is the target language code (change as needed)
+- `--check` is an optional flag to only check for missing keys without prompting for translations
+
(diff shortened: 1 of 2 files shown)


love-linger:
You can find the missing translations in https://github.com/sourcegit-scm/sourcegit/blob/develop/TRANSLATION.md

ramonmelo:
The script is not only for showing the missing tags (the `--check` is an optional argument), but its main purpose is to assist with translation.

Instead of checking each key manually, you can run the script with the desired language, and it will interactively review the missing keys, prompting you for the translation of each one. Once you're done, it automatically saves the translated text back into the target language file, maintaining alphabetical order.

This approach is much simpler than copying and pasting keys, figuring out where to place them, and determining if they are missing or just out of order.

ramonmelo:
Can this be reviewed again?

love-linger:
I'm very sorry for my late reply.

I suggest placing `translate_helper.py` directly in the root directory of the project and writing the relevant usage instructions into the `Translation Status` section of the main `README.md` file.

ramonmelo:
Hi, I've updated the branch with the requested changes. Can you check it again? 

Thanks!

PR #1835, opened 2025-09-30
Auto-stash and restore working tree changes during git bisect operations (AI-generated)

shabbir-sprinting:
This PR implements automatic stashing of local changes when marking commits as "good", "bad", or "skip" during a git bisect operation. If the working directory is not clean, changes are stashed before the bisect step and restored afterward. This improves usability and prevents bisect failures due to uncommitted changes.

This PR was AI-generated.

### DIFF
--- a/src/ViewModels/Repository.cs
+++ b/src/ViewModels/Repository.cs
@@ -1152,8 +1152,17 @@ namespace SourceGit.ViewModels
             using var lockWatcher = _watcher?.Lock();
             IsBisectCommandRunning = true;
 
-            var log = CreateLog($"Bisect({subcmd})");
+            bool hadLocalChanges = LocalChangesCount > 0;
+            string stashName = $"sourcegit-auto-bisect-{DateTime.Now:yyyyMMddHHmmss}";
+            bool stashed = false;
 
+            if (hadLocalChanges) {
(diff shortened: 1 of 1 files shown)


love-linger:
`Bisect` is usually used to locate the commit that caused a known issue. Under normal circumstances, locally uncommitted code should not be included (as it may cause more interference). Therefore, this PR will not be added to SourceGit.

shabbir-sprinting:
> Under normal circumstances, locally uncommitted code should not be included

This is theoretically correct, but as the saying goes, “In theory, there is no difference between theory and practice. In practice, there is.” 🙂

In real-world enterprise repositories, best practices aren’t always followed. For instance, I work with a codebase where I need to maintain local modifications to several Git-tracked files so that the system connects to the correct database for the development environment.

Because of this setup, I’d like to be able to run git bisect without having to stash and reapply these local changes before and after each step.

love-linger:
> For instance, I work with a codebase where I need to maintain local modifications to several Git-tracked files so that the system connects to the correct database for the development environment.

I recommend using the `post-check` hook provided by git to do this.

For example:
```shell
#!/bin/bash

echo "Post-checkout hook executed"
echo "Old HEAD: $1"
echo "New HEAD: $2"
echo "Is branch checkout: $3"
```

PR #1841, opened 2025-10-08
feature: add fast-forward menu for other local branches

cdammanintopix:


love-linger:
SourceGit supports this feature in the context menu of the local branch tree.

image

However, in the context menu of the selected commit in HISTORY page, we only list some of them. This is because:

- The status of a non-current branch and the upstream branch it tracks is not very easy to see in the commit list rather than branch tree. We recommand to do it from the context menu of branch tree.
- The context menu in the commit list is as much as possible related to the selected revision, such as merge/reset, etc.

PR #1842, opened 2025-10-08
Improve logs view

cdammanintopix:
image


cdammanintopix:
@love-linger could you look at this?

love-linger:
- `Views.ViewLogs.OnLoaded` is unnecessary. It is recommanded to move this code to viewmodels.
- `PureText` in `Views.CommandLogContentPresenter` is only used by notifications. In that situation, content should not scroll to the end.
- `Hyperlinks` and `EmailHyperlinks` in `TextEditor` are always disabled in this project currently since in `Dark` theme, they are hard to read.
- The `ViewLogs` window is resizable.

I'll implement this feature myself.

PR #1844, opened 2025-10-09
"Show only my locks" checkbox did not work anymore in the LFS Locks window

mat13109:
The "Show only my locks" checkbox stopped working, even if the username was valid. It was caused by a non-functional bound property. 
So I refactored the `HasValidUserName` property in `LFSLocks.cs` to use a backing field with SetProperty support. The bind now works correctly, and the checkbox can be used again.


mat13109:
Sorry, I will open it again I missed something

PR #1858, opened 2025-10-16
Trim leading/trailing spaces for LFSLock data

mat13109:
I have trimmed the leading/trailing spaces for the `File` and `User` properties in `LFSLock`. These spaces prevented the unlock procedure.


love-linger:
I've pushed my implementation of this.

mat13109:
Thanks for the time. I did not know that LFS could output in json format, this is great. Nicely done!

PR #1863, opened 2025-10-19

jjesus-dev:
Add missing strings.


jjesus-dev:
Hi @love-linger 

I don't get the "Likes" part in the `Text.Configure.CommitMessageTemplate.BuiltinVars` string, can you provide a little more context about what are they being used for? So I can translate them in a more accurate way.

Examples:

A) _I "Like" this file names?_
B) _This file name is "like" the other one?_
C) _You are not even close._

Mingun:
That is a help message for variables, that can be used in text. Obviously, it explains, that `pure_` variables contains the same data, as non-`pure_`, but only filenames, instead of full paths.

jjesus-dev:
> That is a help message for variables, that can be used in text. Obviously, it explains, that `pure_` variables contains the same data, as non-`pure_`, but only filenames, instead of full paths.

So it's like using "as"?
Example: "_Shows/treats `pure` filenames like/as `non-pure` filenames_".

Issue #1912, opened 2025-11-12
[Feature Request] Add merge strategy option for single-branch merge


nelson870708:
**Description:**
Currently, when merging multiple branches, the merge dialog provides an option to choose merge strategies (e.g., --strategy=ours).
However, when merging a single branch, the UI performs the merge immediately without offering any strategy selection.

This inconsistency makes it impossible to use alternative merge strategies (such as ours) for simple one-to-one merges.

**Expected behavior:**
When merging a single branch, show the same merge strategy selection UI (or at least allow specifying a default strategy).

**Motivation:**
Sometimes we need to merge a single branch while always keeping the current branch's content (--strategy=ours), but currently there’s no way to do that in the single-branch merge flow.

**Additional context:**
I reviewed the code and noticed that the multi-branch merge path already supports strategy options.
I’d be happy to help implement or test this if it aligns with the project goals.

PR #1913, opened 2025-11-12
feat: add merge strategy option to merge command and UI

nelson870708:
As I mentioned in #1912, I added merge strategy option to merge UI.

image

### DIFF
--- a/src/Commands/Merge.cs
+++ b/src/Commands/Merge.cs
@@ -5,7 +5,7 @@ namespace SourceGit.Commands
 {
     public class Merge : Command
     {
-        public Merge(string repo, string source, string mode, bool edit)
+        public Merge(string repo, string source, string mode, bool edit, string strategy)
         {
             WorkingDirectory = repo;
             Context = repo;
@@ -14,6 +14,8 @@ namespace SourceGit.Commands
             var builder = new StringBuilder();
             builder.Append("merge --progress ");
             builder.Append(edit ? "--edit " : "--no-edit ");
+            if (!string.IsNullOrEmpty(strategy))
+                builder.Append("--strategy=").Append(strategy).Append(' ');
             builder.Append(source);
             builder.Append(' ');
             builder.Append(mode);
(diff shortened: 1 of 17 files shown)


love-linger:
I'm sorry. I've revert this PR manually because the `octopus` is not suitable for merging two heads.

love-linger:
And for two-heads-merging, the default `ort` merge-strategy is considered more predictable and consistent between your local machine and `GitHub`.

I think for two-heads-merging, you may need `--strategy-option=ours|theirs|..` with the default `ort` merge-strategy instead of choosing another merge-strategy

nelson870708:
Thanks for the clarification!

Regarding `octopus`:  
I understand that it’s primarily intended for merging three or more heads, but technically it can still handle two heads without issues. My original intention was simply to reuse the existing multi-head merge code path to keep the implementation simpler and reduce future maintenance cost. **If using `octopus` for two-head merges is not appropriate for the project, I’m totally fine removing that part.**

About merge **strategy** vs **strategy-option**:  
They are related but conceptually different. A quick example:

- **Merge strategy** (`--strategy=`) chooses the merge algorithm:
  ```
  git merge --strategy=recursive
  git merge --strategy=ort
  git merge --strategy=ours
  ```

- **Merge strategy options** (`--strategy-option=`) modify the behavior *within* a chosen strategy.  
  For example:
  ```
  git merge --strategy=ort --strategy-option=ours
  ```
  In this case, the strategy is still `ort`, not `ours`—the option only tells `ort` how to resolve conflicts.  
  This distinction is subtle but important in real workflows.

A concrete example from my workflow as a research engineer:  
I often create an experimental branch for testing different algorithms or configurations. After an experiment is finished, I want to merge it back into the `develop` branch **without letting the experimental changes overwrite anything in `develop`**—but I also don’t want to lose the experimental commits, since I may need to revisit them later to check certain configurations.
Using `--strategy=ours` perfectly solves this scenario: it lets me merge the branch while keeping `develop` untouched and still preserving the experimental commit history.

The main motivation of the PR is to make the single-branch merge flow consistent with the multi-branch flow—so users can choose the appropriate strategy or strategy-options (like always keeping the current branch) even when merging just one branch.

If you feel a different approach fits the project better, I’m happy to adjust the PR accordingly or help refine the implementation.

Thanks again for the feedback and for reviewing the PR!

PR #1952, opened 2025-12-03
[BUG] Fix: AutoScrollToFirstChange() was not working reliably for Side-By-Side Diff

goran-w:
Both of the `SingleSideTextDiffPresenter` instances need to be scrolled/synced in `AutoScrollToFirstChange()`, for the auto-scrolling to work reliably. (To see where it failed, open a `Side-By-Side Diff` where the first change-block would be out-of-view, then toggle `Show All Lines` - when enabling the latter, the first change-block would NOT be correctly scrolled-to.)

As a bonus, a bit of cleanup and simplification was also done in related code:
* Added private helper method GotoChange(), to remove redundant code.
* Made `SyncScrollOffset()` virtual, so the `Goto<...>Change()` methods could be removed.
* In `DiffContext.CheckSettings()`, setting the `Context` is now made on a single (less redundant) line, just like in the `DiffContext.UseSideBySide` setter.

### DIFF
--- a/src/ViewModels/DiffContext.cs
+++ b/src/ViewModels/DiffContext.cs
@@ -136,8 +136,7 @@ namespace SourceGit.ViewModels
 
                 if (ctx.IsSideBySide() != UseSideBySide)
                 {
-                    ctx = ctx.SwitchMode();
-                    Content = ctx;
+                    Content = ctx.SwitchMode();
                 }
             }
         }
--- a/src/Views/TextDiffView.axaml.cs
+++ b/src/Views/TextDiffView.axaml.cs
@@ -487,44 +487,38 @@ namespace SourceGit.Views
         {
         }
 
-        public virtual void GotoFirstChange()
+        protected virtual void SyncScrollOffset()
         {
-            var first = BlockNavigation.GotoFirst();
-            if (first != null)
+        }
+
+        private void GotoChange(ViewModels.BlockNavigation.Block block)
+        {
+            if (block != null)
             {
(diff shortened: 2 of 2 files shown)


love-linger:
`SyncScrollOffset()` should not be called for `Combined` diff mode

love-linger:
I've pushed my fix for this issue

goran-w:
> I've pushed my fix for this issue

Great - looks good!

Seems like the `SyncScrollOffset()` method was no longer needed, at all...

A possible future improvement could be to jump to the **_current_** change-block (instead of back to the first one) when toggling `Show All Lines` and `Side-By-Side Diff`, since the set of changes are actually the same (only the visualization changes).

It's all about not losing context, when toggling these options - which I often do specifically to get another view of the very same change-block I'm looking at...

goran-w:
Excellent work on keeping the current change-block when toggling `Show All Lines`!

Could we also have this when toggling `Side-By-Side Diff`?

For that, it would probably be (more or less) sufficient to make the following changes :
* Make the `CombinedTextDiff` and `TwoSideTextDiff` constructors take their `previous` arg as the base-class `TextDiffContext` (instead of as their own specific sub-classes).
* Make the `SwitchMode()` methods pass the current (`this`) `TextDiffContext` as the `previous` arg to the constructor of the switched-to `TextDiffContext` sub-class.

goran-w:
Marvelous, current change-block is now kept also when toggling `Side-By-Side Diff` - thank you! 👍

PR #1956, opened 2025-12-05
feature: add support for additional flags in push command

equinoxmatt:
A little bug bear of mine. Sometimes for good reason, I need to skip the verify command. Previously I always had to revert to the CLI, I would prefer to be able to do this within sourcegit.

This allows users to pass custom git flags (e.g., --no-verify, --dry-run) to the push command through a new optional text field in the push dialog.

image

image

### DIFF
--- a/src/Commands/Push.cs
+++ b/src/Commands/Push.cs
@@ -4,7 +4,7 @@ namespace SourceGit.Commands
 {
     public class Push : Command
     {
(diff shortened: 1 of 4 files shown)


love-linger:
I'm sorry that this PR will not be merged.

I recommend that you define a custom push operation (using the Custom Action function).

PR #1958, opened 2025-12-06
feat: implement Windows 11 Snap Layouts support for custom caption buttons (#1957)

### LINKED ISSUE #1957: [BUG] SourceGit doesn't fully support Windows 11 'Snap Layouts'
When hovering over an application's Maximize/Restore button (on its title bar) in Windows 11, a dropdown/popup menu of [Snap Layouts](https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/ui/apply-snap-layout-menu) should appear.

This does not currently work in SourceGit, which is using Avalonia 11.3.9 since v2025.39.

However, this feature (https://github.com/AvaloniaUI/Avalonia/issues/11409, https://github.com/AvaloniaUI/Avalonia/pull/17380) SHOULD be supported by Avalonia since [11.3.0](https://github.com/AvaloniaUI/Avalonia/releases/tag/11.3.0).

A related feature (https://github.com/AvaloniaUI/Avalonia/pull/19922) merged into master was confirmed to still work along with https://github.com/AvaloniaUI/Avalonia/issues/11409 (see [this comment](https://github.com/AvaloniaUI/Avalonia/pull/19922#issuecomment-3503150591)). This feature was, backported into [11.3.9](https://github.com/AvaloniaUI/Avalonia/releases/tag/11.3.9) via PR https://github.com/AvaloniaUI/Avalonia/pull/20076.

Could it be that SourceGit is overriding something in the handling of main window titlebar / buttons that breaks this? (The custom Window-subclass `ChromelessWindow` might be a culprit here, somehow?)

-----

NOTE: Basic window-snapping (when moving the window and approaching snap-zones at screen edges) works fine, sort-of, but there's actually another bug found there:
* If the SourceGit main window is moved to approach a neighboring (multi-)monitor wich has a different screen resolution and DPI / scaling, the window gets resized / rescaled (to fit the other monitor's resolution) when more than half of the window is already on that other monitor's screen. So far, so good - but when moving even closer to the first monitor's edge, if the mouse pointer enters a snap zone and we release it there, the window (after being correctly resized / rescaled back to the resolution of the first monitor) does one of two things (which are both wrong):
  * Either (if initially smaller than the snap-zone), it gets snap-docked to the correct snap-corner but its final visible size is NOT correctly updated, leaving it in a rather strange situation where the window is visually smaller than the snap-zone! However, its (invisible) client-area still fills the whole snap-zone, which means we can't click on any window behind it (within the snap-zone area), since SourceGit still receives these clicks! (If the window is then simply minimized and restored, however, it suddenly "owns up" to fill the entire snap-zone correctly...)
  * Or  (if initially larger than the snap-zone), it simply gets resized to a size smaller than the snap-zone and left "undocked" (somewhere near, but not necessarily within, the snap-zone).
* Tip: to reproduce these problematic situations, make sure to grab the main window in a position far from the monitor edge that borders another monitor, so that more than half of the window ends up on the other monitor before we enter the snap-zone. These issues only show up when combining the zone-snapping with the resize / rescale due to resolution differences - snapping without simultaneous rescaling seems to work fine!
* These issues are not seen with other applications' main windows (like Google Chrome) - they get resized / rescaled the same way but are still snap-docked correctly (regardless of being initially smaller or larger than the snap-zone).

gadfly3173:
Fixes #1957

Tested on Windows 10 & 11, have not test on Linux or macOS. Did not test or fix the bug about multi monitors.

Edit: Tested on Deepin V25 (X11), worked fine as before.


love-linger:

- `Win32Properties.AddWndProcHookCallback` will only handle window border hittest event.
- Use `AvaloniaUI`'s built-in `Win32Properties.NonClientHitTestResult` property.

PR #1969, opened 2025-12-11
feature: add fixup head into parent

cdammanintopix:


love-linger:
So what's the different between squashing HEAD into parent and fixuping HEAD into parent?

cdammanintopix:
The final commit message is the difference.

- Squash: add the commit message of both commits
- Fixup: only keep the commit message of the parent

 I personally often use fixup instead of squash, so that the message of the final commit does not change

cdammanintopix:
I just pushed https://github.com/sourcegit-scm/sourcegit/pull/1969/commits/c5fe33a6b3048f259fae4450e8edae800c567304 regarding the same concern

PR #1981, opened 2025-12-16
update nuget package textmate to 2.0

heartacker:


love-linger:
This is a major version change of `TextMateSharp`. And this project requires `TextMateSharp` because that `AvaloniaEdit` uses it.

I'd like to wait for that `AvaloniaEdit` upgrade its dependency.

love-linger:
I see that the `AvaloniaEdit` project has a branch that tries to upgrade `TextMateSharp` and there is indeed an issue of API incompatibility.

So, I'll close this PR and wait for the next version of `AvaloniaEdit`

love-linger:
image

PR #1983, opened 2025-12-16
Default tracking branch for worktrees

AndreasBroager:
When the user toggles the checkbox for "Set tracking branch" in the worktree dialog, then the default tracking branch is set as the same as the selected branch.


love-linger:

PR #1990, opened 2025-12-18
add background fetch to log + background fetch respects git defaults

mithom:
* adds a log entry for background fetches as they are git commands being run by user configuration of the application.
* in background fetches there is no possibility to choose between with or without tags. This needs to either be a user configuration, or use the 'tagOpt' git configuration (which it does by default when not specifying an option). To achieve this we need a third value for noTags, namely default. i introduced this through a nullable value with null being the 3th value.

possible changes to this PR:
* split the commit into 2 commits/PR's (but since it's so small i thought it wouldn't be an issue)
* introduce an enum instead of using null as a value
* make a repo or global configuration for the noTags in the background fetch.

if you want any of the mentioned changes (or any other), please let me know.

### DIFF
--- a/src/Commands/Fetch.cs
+++ b/src/Commands/Fetch.cs
@@ -5,7 +5,7 @@ namespace SourceGit.Commands
 {
     public class Fetch : Command
     {
-        public Fetch(string repo, string remote, bool noTags, bool force)
+        public Fetch(string repo, string remote, bool? noTags, bool force)
         {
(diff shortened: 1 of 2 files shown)


mithom:
the failed task is due to a timeout on the fetching of 'actions/download-artifact@v4'
and is not specific to this PR

love-linger:

PR #2003, opened 2025-12-23
Enhance CommitMessageToolBox

massimopaganigh:
Increase the padding of the three buttons to make them easier to press and prevent two context menus from reopening.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 .vs/
 .vscode/
 .idea/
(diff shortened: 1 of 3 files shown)


massimopaganigh:
This PR includes a change to `.gitignore` to ignore files related to the DevExpress CodeRush extension. Consider whether it may be useful for other individuals as well.

love-linger:

Issue #1998, opened 2025-12-20
Very slow staging files on big repositories on macOS


kunitoki:
I'm experiencing very slow (sometimes it takes even 20 seconds) staging of changed files on big repositories on macOS. Other git clients or even git from command line don't exhibit the same behaviour, is it something you are experiencing ?

love-linger:
Can not re-produce on my macOS device. I just tested it in my Unreal Engine repo.

kunitoki:
My repo is really big, once i have generated and compiled the final product, it's around 40GB in size and can contain over 200k files

AlanIWBFT:
`git status` equals to `git status --untracked-files=normal`. Afaik, `--untracked-files=all` can be very slow on Windows (~5 to ~20s on UE5 repo, seems depending on how much you touched). On Linux it is much faster (~1s consistenly). It just seems like Git's untracked cache is not helping so much when `-uall` is added. Since `-uall` is required to accurately report what has been changed (so that you see each untracked file), I see this as a Git issue thus there's no easy solution. On Windows I just comment out all code that uses `-uall` in SourceGit and only manually look for untracked files when I need.

love-linger:
> On Windows I just comment out all code that uses -uall in SourceGit and only manually look for untracked files when I need.

Actually, after I tested it, I don't feel any changes.

love-linger:
I've pushed a commit that tries to improve stage/unstage/discard performance. You can test it.

AlanIWBFT:
> Actually, after I tested it, I don't feel any changes.

The effect is the most pronounced when Visual Studio is running:
```
PS E:\ue5-main> Measure-Command { git status --untracked-files=all }

Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 11
Milliseconds      : 41
Ticks             : 110412224
TotalDays         : 0.000127791925925926
TotalHours        : 0.00306700622222222
TotalMinutes      : 0.184020373333333
TotalSeconds      : 11.0412224
TotalMilliseconds : 11041.2224

PS E:\ue5-main> Measure-Command { git status }

Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 469
Ticks             : 4699134
TotalDays         : 5.4388125E-06
TotalHours        : 0.0001305315
TotalMinutes      : 0.00783189
TotalSeconds      : 0.4699134
TotalMilliseconds : 469.9134
```

[...]

PR #2012, opened 2025-12-26
enhance: implement fast path for repository local change status update when `core.fsmonitor` and `core.untrackedcache` are enabled

AlanIWBFT:
Finally this experiment has come to a stage where it's reviewable. It is still very early. Opinions are welcome. Binaries for testing (win-x64 only) can be downloaded here: https://github.com/AlanIWBFT/sourcegit/releases

### Problem it aims to solve: https://github.com/sourcegit-scm/sourcegit/issues/1998
To accurately report and let the user pick each individual untracked file, SourceGit uses `--untracked-files=all` (`-uall`) when doing `QueryLocalChanges`, which is unfortunately very slow on large repositories. `core.fsmonitor` and `core.untrackedcache` cannot help this case if `-uall` is specified as it makes the untracked cache bypassed, yielding no performance improvement.
This pull request detects if `core.fsmonitor` and `core.untrackedcache` are enabled and uses a two pass algorithm: it runs a normal `git status` first to retrive a list of untracked directories (which is fast as untracked cache is functioning here), then run `git status -uall` only on these dirs. The time it takes to traverse these dirs are much faster than running `-uall` on the entire repository.

To use this experimental feature, the user must deliberately enable `core.fsmonitor` and `core.untrackedcache`, or not setting `core.untrackedcache` but setting `feature.manyfiles` to `true`.

### Components
First, to robustly detect if the fast path is possible, `Commands.Config.GetBoolAsync` is added which leverages git itself to canonicalize possible weird values in config files, like "000" will be intepreted as "false", "010" -> "true".
Every time you open a repository the above detection is performed. If they are enabled, a normal, index-locking 'git status' will be run to populate/update the untracked cache. The first time will be slow as the cache needs to be populated entirely from empty. Subsequent runs will be much faster, in fact, negligible. (~0.4s on UE5-main). Thus the slow-down on UI start-up isn't very noticable expect for the first time. Since SourceGit always runs `git status` with `--no-optional-locks` which means the index will not be modified, the untracked cache thus cannot be updated either, once SourceGit is up. So we need a place to insert the update, and it is scheduled when you open the repository, as said above.

### Performance results
On large repos like UE5, a consistent ~5x - ~6x speedup is observed whenever a status update is needed and the responsiveness is greatly improved.

### Other fixes
As we've discussed in https://github.com/sourcegit-scm/sourcegit/issues/1998#issuecomment-3692145166, `Stash.PushAsync` `bool includeUntracked`'s default value has been changed to `false`. The intention is to let git handle the untracked instead of manually stash them (which was performed with `--included-untracked`, very slow (2x slower than the slow status update)).

### DIFF
--- a/src/Commands/Config.cs
+++ b/src/Commands/Config.cs
@@ -74,6 +74,27 @@ namespace SourceGit.Commands
             return rs.StdOut.Trim();
         }
 
+        // Get config value with type canonicalization
+        // Weird values will be converted by git, like "000" -> "false", "010" -> "true"
+        // git will report bad values like "fatal: bad boolean config value 'kkk' for 'core.untrackedcache'"
+        public async Task<bool?> GetBoolAsync(string key)
+        {
+            Args = $"config get --bool {key}";
+
+            var rs = await ReadToEndAsync().ConfigureAwait(false);
+            var stdout = rs.StdOut.Trim();
+            switch (rs.StdOut.Trim())
+            {
+                case "true":
+                    return true;
+                case "false":
+                    return false;
+                default:
+                    // Illegal values behave as if they are not set
+                    return null;
+            }
+        }
+
         public async Task<bool> SetAsync(string key, string value, bool allowEmpty = false)
         {
             var scope = _isLocal ? "--local" : "--global";
--- a/src/Commands/QueryLocalChanges.cs
+++ b/src/Commands/QueryLocalChanges.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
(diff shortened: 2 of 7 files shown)


AlanIWBFT:
Oops, looks like I got the branch wrong and pulled in lots of commits. Fixing

love-linger:
I don't like to do this, but I prefer to predict the changes of ‘Local Changes’ after the ‘add/reset’ operation is successful, so that even if there is something wrong, such as we cannot predict the renames, the list will be updated to the correct state after the background ‘git status’ command returns.

AlanIWBFT:
Maintaining consistency can be hard. Afaik predicting what's changed is how UGit works. Since the background operation is slow, UGit needs to run in the background (minimized to tray) until the operation is done. If you try to stage something, it shows its staged, then if you exit UGit immediately and reopen it, you see your changes are not staged.
This pull request, while seems 'not elegant' at first glance, ensures the status you see is always real, i.e. no consistency issue to battle in the first place.

AlanIWBFT:
The real consistency issue emerges, when the user performs multiple ops in serial. Its like UE network prediction and rollback - if git status fails to catch up timely, you will need to design an algorithm to determine which user ops have failed and rollback to the specific point. If you don't allow such operation queues, the status update will be as slow as before once your user launched the second operation.
edit: example ops of such dependenices are: the user selects only _some_ of the files in the shown status list to stage/unstage/dicard/stash
edit2: the most problematic case is when the user is mixing these ops while also changing files in the working copy - without a correct git status it isnt easy to do the predictions right

Issue #1950, opened 2025-12-02
Discard not working


MrJacquers:
Discard isn't working for me in the diff view. I've tried clicking the popup and ctrl-d, but nothing happens. Stage works though.

love-linger:
Can not re-produce this issue. Can you provide a minimal git repo?

MrJacquers:
It's happening on my work laptop and I can't share that repo or example. I'll have to check on my personal laptop and see if it happens there. Or maybe clone an open-source repo and check using that.

goran-w:
Does this only happen on specific files, or everywhere? Is it the same in a new dummy repo with some simple changes?

Dl1MA:
I have same problem, while try to discard last rows in the `.*sql` file. It's not work. But at the middle or begin of file - discard fine. (Version `2025.39`)

love-linger:
@Dl1MA Can you provide a minimal diff file? And what's your settings in diff view (--ignore-crlf-at-eol, --ignore-spaces, etc.)

CosmoLau:
@love-linger 我发现当文件属性设置为 `只读` 时，新增这个只读文件，就无法丢弃，会显示如下报错：

```text
Failed to discard changes. Reason: Access to the path 'J:\CosmoLau\Workspace\other\discard-demo\123456 - 副本.docx' is denied.
```

使用 `git clean -df` 命令行是可以成功丢弃的

love-linger:
@sina-hide has pushed a commit that tries to fix this issue.

> 我发现当文件属性设置为 只读 时，新增这个只读文件，就无法丢弃，会显示如下报错：

This is not a issue that belongs to `SourceGit` or even `git`

I'll close this issue first. If this issue still exists, you can open it at anytime.

PR #2027, opened 2026-01-04
fix: last line change may disable hunk operation in two-side diff

sina-hide:
If the last line is deleted or added, it was impossible to stage, unstage or discard the last hunk in a two-side diff view, if the hunk was selected on the side where the last line is missing (that's the left side for added lines and the right side for deleted lines).

Since the button(s) for hunk operations are on the right side of the diff view, it was easier to trigger this bug for a deleted last line.

This bug was introduced in commit 6511d15c01 while rewriting the text diff views.

This fixes the bug mentioned in https://github.com/sourcegit-scm/sourcegit/issues/1950#issuecomment-3630580428.

### DIFF
--- a/src/ViewModels/TextDiffContext.cs
+++ b/src/ViewModels/TextDiffContext.cs
@@ -191,18 +191,24 @@ namespace SourceGit.ViewModels
             _option = option;
             _data = diff;
 
-            foreach (var line in diff.Lines)
+            var sourceIndices = new int[diff.Lines.Count];
+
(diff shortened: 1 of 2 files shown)


love-linger:
- This code is very hard to read.
- Since the `New` side always has the same line with `Old` side for modified lines, the `sourceIndices[New.Count]` will always override the `sourceIndices[Old.Count]`

And this issue does not exists when trying to stage/discard/unstage with selected contents (not auto-detected hunk) in one side. 

I'll push my commit to fix this issue

PR #2049, opened 2026-01-15
feat: expand and collapse all buttons

massimopaganigh:
feat

### DIFF
--- a/src/Resources/Locales/en_US.axaml
+++ b/src/Resources/Locales/en_US.axaml
@@ -708,12 +708,14 @@
   <x:String x:Key="Text.Repository.CleanTips" xml:space="preserve">Run `git gc` command for this repository.</x:String>
   <x:String x:Key="Text.Repository.ClearAllCommitsFilter" xml:space="preserve">Clear all</x:String>
   <x:String x:Key="Text.Repository.ClearStashes" xml:space="preserve">Clear</x:String>
+  <x:String x:Key="Text.Repository.CollapseAll" xml:space="preserve">Collapse All</x:String>
   <x:String x:Key="Text.Repository.Configure" xml:space="preserve">Configure this repository</x:String>
   <x:String x:Key="Text.Repository.Continue" xml:space="preserve">CONTINUE</x:String>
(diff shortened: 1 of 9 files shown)


massimopaganigh:
I'm honestly quite torn about what the best position for those two buttons might be.

love-linger:
I think this PR is not suitable for `SourceGit`.

- Currently, there are only 5 groups (which is not many), and each group manages completely different things (Branches/Tags/Submodules/Worktrees).
- The layout of the entire left-sidebar has been manually optimized to avoid scrollbars appearing on the entire page, which would cause users to have to scroll the view first when wanting to close or open a certain group.
- The `Expand/Collapse All` operation actually has few application scenarios. Especially `Expand All`, I'm not sure in what situations users would need to expand all groups (as mentioned in the first point, these groups are not managing the same things, and in the second point the vertical space is too small that we need manually layout these groups). Moreover, even for `Collapse All`, I basically never use it because at least the `LOCAL BRANCHES` group will always be displayed by me.

massimopaganigh:
@love-linger So maybe it would make sense to have a specific collapse option for Local Branches. I have several repositories that are structured in different subfolders (e.g., MILESTONE/TYPEOFCHANGE/#ISSUE/#SUBISSUE/...) and there is no real way to tidy them up by collapsing them all with a single click (as in Visual Studio).
image

massimopaganigh:
Now that I think about it, it would also make sense to have a button to focus on the current branch (as in Visual Studio for the current document).

love-linger:
> Now that I think about it, it would also make sense to have a button to focus on the current branch (as in Visual Studio for the current document).

Try this button.
image

massimopaganigh:
> @love-linger So maybe it would make sense to have a specific collapse option for Local Branches. I have several repositories that are structured in different subfolders (e.g., MILESTONE/TYPEOFCHANGE/#ISSUE/#SUBISSUE/...) and there is no real way to tidy them up by collapsing them all with a single click (as in Visual Studio). image

? Should I try implementing it this way? Or can I close the PR here?

love-linger:
I'm very sorry. I think you can just close this PR

massimopaganigh:
@love-linger Aight, nw ^^

PR #2088, opened 2026-02-01
feature: Move cache to its own directory to follow XDG standards.

flipper:
Cache should not be in the same directory as the configuration. It mostly only applies to linux.

See here for what each folder maps to https://xdg-net.github.io/Xdg.Directories/docs/defaults.html

### DIFF
--- a/src/Models/AvatarManager.cs
+++ b/src/Models/AvatarManager.cs
@@ -44,7 +44,7 @@ namespace SourceGit.Models
 
         public void Start()
         {
-            _storePath = Path.Combine(Native.OS.DataDir, "avatars");
+            _storePath = Path.Combine(Native.OS.CacheDir, "avatars");
             if (!Directory.Exists(_storePath))
                 Directory.CreateDirectory(_storePath);
 
--- a/src/Models/IpcChannel.cs
+++ b/src/Models/IpcChannel.cs
@@ -16,7 +16,7 @@ namespace SourceGit.Models
         {
             try
             {
-                _singletonLock = File.Open(Path.Combine(Native.OS.DataDir, "process.lock"), FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.None);
+                _singletonLock = File.Open(Path.Combine(Native.OS.RuntimeDir, "process.lock"), FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.None);
                 IsFirstInstance = true;
                 _server = new NamedPipeServerStream(
                     "SourceGitIPCChannel" + Environment.UserName,
--- a/src/Native/OS.cs
+++ b/src/Native/OS.cs
@@ -33,6 +33,18 @@ namespace SourceGit.Native
             private set;
         } = string.Empty;
 
+        public static string CacheDir
+        {
+            get;
+            private set;
+        } = string.Empty;
+
+        public static string RuntimeDir
+        {
+            get;
+            private set;
+        } = string.Empty;
+
         public static string GitExecutable
         {
             get => _gitExecutable;
@@ -150,14 +162,41 @@ namespace SourceGit.Native
                 }
             }
 
-            var osAppDataDir = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
-            if (string.IsNullOrEmpty(osAppDataDir))
-                DataDir = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".sourcegit");
-            else
-                DataDir = Path.Combine(osAppDataDir, "SourceGit");
+            DataDir = Path.Combine(Xdg.Directories.BaseDirectory.ConfigHome, "SourceGit");
+            CacheDir = Path.Combine(Xdg.Directories.BaseDirectory.CacheHome, "SourceGit");
+            RuntimeDir = Path.Combine(Xdg.Directories.BaseDirectory.RuntimeDir, "SourceGit");
 
             if (!Directory.Exists(DataDir))
+            {
                 Directory.CreateDirectory(DataDir);
+            }
+
+            if (!Directory.Exists(CacheDir))
+            {
+                Directory.CreateDirectory(CacheDir);
+            }
+
+            if (!Directory.Exists(RuntimeDir))
+            {
+                Directory.CreateDirectory(RuntimeDir);
+            }
+
+            var osAppDataDir = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
+            var oldDir = string.IsNullOrEmpty(osAppDataDir) ?
+                Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".sourcegit") :
+                Path.Combine(osAppDataDir, "SourceGit");
+
+
+            if (Directory.Exists(oldDir) && oldDir != DataDir)
+            {
+                foreach (var file in Directory.GetFiles(oldDir))
+                {
+                    var fileName = Path.GetFileName(file);
+                    File.Move(file, Path.Combine(DataDir, fileName));
+                }
+
+                Directory.Delete(oldDir, true);
+            }
         }
 
         public static void SetupExternalTools()
--- a/src/SourceGit.csproj
+++ b/src/SourceGit.csproj
@@ -63,6 +63,7 @@
     <PackageReference Include="Pfim" Version="0.11.4" />
     <PackageReference Include="TextMateSharp" Version="2.0.2" />
     <PackageReference Include="TextMateSharp.Grammars" Version="2.0.2" />
+    <PackageReference Include="Xdg.Directories" Version="0.1.2" />
   </ItemGroup>
 
   <ItemGroup>


love-linger:
This app supports `Portable Mode` for Windows packages and Linux AppImages. Therefore, we need to store all files generated by `SourceGit` in a single directory.

love-linger:
Moreover, I believe that doing so will bring another benefit: when we want to delete `SourceGit`, we can be certain that all files generated by `SourceGit` can be easily deleted without any residues.

flipper:
> This app supports `Portable Mode` for Windows packages and Linux AppImages. Therefore, we need to store all files generated by `SourceGit` in a single directory.

Portable mode should still be supported is there anything I missed?

flipper:
> Moreover, I believe that doing so will bring another benefit: when we want to delete `SourceGit`, we can be certain that all files generated by `SourceGit` can be easily deleted without any residues.

Correct. When running on Mac or Windows everything is in a single directory.

But on Linux it is two folders. The cache does not belong next to configuration files.

love-linger:
I've pushed a commit that moves data from `$XDG_CONFIG_HOME/SourceGit` or `~/.sourcegit` to `$XDG_DATA_HOME/SourceGit` on Linux

flipper:
> I've pushed a commit that moves data from `$XDG_CONFIG_HOME/SourceGit` or `~/.sourcegit` to `$XDG_DATA_HOME/SourceGit` on Linux

What else does source git store? It should be in .config if it's only configuration

love-linger:
The files generated by `SourceGit` do not strictly distinguish between configuration files, cache data, solidified data, runtime process-related files, etc. All files generated by it are uniformly regarded as program data.

flipper:
> The files generated by `SourceGit` do not strictly distinguish between configuration files, cache data, solidified data, runtime process-related files, etc. All files generated by it are uniformly regarded as program data.

On Windows? Correct 

But on Linux config files are in .config/appname
And cache files are in .cache/appname

flipper:
I saw your commits you did that was basically what this PR is doing. But you are hardcoding the values of XDG dirs. They are env vars. Not static values.

love-linger:
> But on Linux config files are in .config/appname

In my opinion, the **TRUE** configuration files (those that need to be placed in the `/etc/` or `$XDG_CONFIG_HOME` directories) are all files that require manual management (direct modification) by users. They are mostly applicable to system basic components and services. For GUI programs, even configuration settings should try to avoid direct editing by users (as they may fail to load due to format issues, etc.), and instead provide an interface for modification.

Many modern softwares directly use `~/.{app_name}` to store all its generated files, including configurations and process-related data. For example: even `git` stores its global configuration directly in `~/.gitconfig`.

> I saw your commits you did that was basically what this PR is doing

Actually, they are not the same.

- Only changes related to the Linux platform are involved; other platforms remain consistent with previous behaviors.
- It does not distinguish between `Config`, `Cache`, and `Process-related` directories. P.S. this PR has a serious issue: when portable mode is detected, both `CacheDir` and `RuntimeDir` are empty values.
- <s>Ensure that users of older versions can upgrade without noticing and retain their previous data.</s>

> But you are hardcoding the values of XDG dirs. They are env vars. Not static values.

Yes, this is intentional. I hope this path is preferably fixed and easy for users to find.

flipper:
> > But on Linux config files are in .config/appname
> 
> In my opinion, the **TRUE** configuration files (those that need to be placed in the `/etc/` or `$XDG_CONFIG_HOME` directories) are all files that require manual management (direct modification) by users. They are mostly applicable to system basic components and services. For GUI programs, even configuration settings should try to avoid direct editing by users (as they may fail to load due to format issues, etc.), and instead provide an interface for modification.
> 

Ideally they should be edited by the GUI but it has nothing to do where the files are actually stored. It just following a standard pattern.

> Many modern softwares directly use `~/.{app_name}` to store all its generated files, including configurations and process-related data. For example: even `git` stores its global configuration directly in `~/.gitconfig`.
> 

No. Modern software does not put itself in ~/.app_name. Even git uses ~/.config/git/config
Process-related data has always been in /run/user/1000/(whatever)

> > I saw your commits you did that was basically what this PR is doing
> 
> Actually, they are not the same.
> 
>     * Only changes related to the Linux platform are involved; other platforms remain consistent with previous behaviors.
> 
>     * It does not distinguish between `Config`, `Cache`, and `Process-related` directories. P.S. this PR has a serious issue: when portable mode is detected, both `CacheDir` and `RuntimeDir` are empty values.
> 
>     * Ensure that users of older versions can upgrade without noticing and retain their previous data.
> 
> 
> > But you are hardcoding the values of XDG dirs. They are env vars. Not static values.
> 
> Yes, this is intentional. I hope this path is preferably fixed and easy for users to find.

There is no point of hardcoding the paths when there is ENV vars to use. If an advanced user has their XDG_CONFIG_HOME set to /tmp. The app should follow it.

Portable mode is for Windows only. CacheDir is set to application data and same with runtime dir. See the table here:
https://xdg-net.github.io/Xdg.Directories/docs/defaults.html

love-linger:
Just run `ls -a` under your home dir. For example, here's mine:

```
.ssh/
.vscode/
.gnupg/
.dotnet/
.gitconfig
```

love-linger:
> There is no point of hardcoding the paths when there is ENV vars to use. If an advanced user has their XDG_CONFIG_HOME set to /tmp. The app should follow it.

Now, `SourceGit` clearly stores its data into `~/.sourcegit` dir (it has nothing to do with `XDG_XXX_HOME` envs). All the files under this folder is datas (that means you can use `SourceGit` without any configuration file). Event `preference.json` is just a file that stores current state of `SourceGit`, it's not a configuration file, remember.

flipper:
> Just run `ls -a` under your home dir. For example, here's mine:
> 
> ```
> .ssh/
> .vscode/
> .gnupg/
> .dotnet/
> .gitconfig
> ```

And? Applications are adapting the XDG standard. If you google how to edit your gitconfig everyone says edit ~/.gitconfig. But git actually supports reading from ~/.config/git/config. 
I do not see the reason to follow old patterns.

You can read more about the standard here: https://wiki.archlinux.org/title/XDG_Base_Directory

flipper:
> > There is no point of hardcoding the paths when there is ENV vars to use. If an advanced user has their XDG_CONFIG_HOME set to /tmp. The app should follow it.
> 
> Now, `SourceGit` clearly stores its data into `~/.sourcegit` dir (it has nothing to do with `XDG_XXX_HOME` envs). All the files under this folder is datas (that means you can use `SourceGit` without any configuration file). Event `preference.json` is just a file that stores current state of `SourceGit`, it's not a configuration file, remember.

Okay. If you think it is more of a data file. Then that file should be in $XDG_DATA_HOME/SourceGit. But do not assume .local/share in the code. It is a env var for a reason.

Avatars is not a data file. It is a cache. So it should be in $XDG_CACHE_HOME/SourceGit/avatars/

jernejp21:
Having a `~/.sourcegit` is a legacy thing. Programs shouldn't be using this any more.

But also environment variables only shouldn't be used. Some systems still don't have these variables set by default, so a hard coded path should be used. Please see [Freedesktop.org](https://specifications.freedesktop.org/basedir/latest/)

mpeter50:
@love-linger 
> In my opinion, the TRUE configuration files (those that need to be placed in the /etc/ or $XDG_CONFIG_HOME directories) are all files that require manual management (direct modification) by users. They are mostly applicable to system basic components and services. For GUI programs, even configuration settings should try to avoid direct editing by users (as they may fail to load due to format issues, etc.), and instead provide an interface for modification.

As a linux user, config files should not be _required_ to be edited manually. Programs dont hold such an assumption. Plenty of them store their user settings in .config/appname. Of course they are not required to automatically reload the configuration when the file is edited, generally there is no such expectation. One of the benefits of this is the maintainability of synchronizing settings between computers, with a dotfiles manager program or something else.

> Many modern softwares directly use ~/.{app_name} to store all its generated files, including configurations and process-related data.

That is generally discouraged and frowned upon, because it clutters the home directory. Mostly very old programs do that, that were created before there was any standardization, and didnt bother to support the standard config locations.
Recenlty even Firefox was updated to place its files somewhere in ~/.local for new installations (and anyone who manually moves it there).

> For example: even git stores its global configuration directly in ~/.gitconfig.

Debian does not necessarily store that file there. According to the manual, it prefers `$XDG_CONFIG_HOME/git/config` if it exists, and thats what it is doing on all my computers too.

```
FILES
       By default, git config will read configuration options from multiple files:

       $(prefix)/etc/gitconfig
           System-wide configuration file.

       $XDG_CONFIG_HOME/git/config, ~/.gitconfig
           User-specific configuration files. When the XDG_CONFIG_HOME environment variable is not set or empty, $HOME/.config/ is used as $XDG_CONFIG_HOME.

           These are also called "global" configuration files. If both files exist, both files are read in the order given above.
```

> Just run ls -a under your home dir. For example, here's mine:

The issue is not discoverability, but clutter. The root of the home direcotry is for the files of the user, not random programs, and if programs put their files there users wont find their own things among all the litter. Look into your `%AppData%` folder, and imagine if all that was in your home folder on windows. It would be horrible.

Please restore the old behavior. Storing all files in `~/.local/share/sourcegit/` is not that big of a deal, but keeping and moving them to `~/.sourcegit/` is a significant downgrade. I'll probably not update until sourcegit is littering the home directory. Please understand that this is not some meaningless thing, it really benefits the users when the proper directories are used.
If you need help, please let me know.

mpeter50:
@flipper

> Portable mode is for Windows only. 

I dont think portable mode needs to be windows only. Why cant portable mode use the portable config directory, and then the common cache and runtime dirs?

flipper:
> @love-linger
> 
> > In my opinion, the TRUE configuration files (those that need to be placed in the /etc/ or $XDG_CONFIG_HOME directories) are all files that require manual management (direct modification) by users. They are mostly applicable to system basic components and services. For GUI programs, even configuration settings should try to avoid direct editing by users (as they may fail to load due to format issues, etc.), and instead provide an interface for modification.
> 
> As a linux user, config files should not be _required_ to be edited manually. Programs dont hold such an assumption. Plenty of them store their user settings in .config/appname. Of course they are not required to automatically reload the configuration when the file is edited, generally there is no such expectation. One of the benefits of this is the maintainability of synchronizing settings between computers, with a dotfiles manager program or something else.
> 
> > Many modern softwares directly use ~/.{app_name} to store all its generated files, including configurations and process-related data.
> 
> That is generally discouraged and frowned upon, because it clutters the home directory. Mostly very old programs do that, that were created before there was any standardization, and didnt bother to support the standard config locations. Recenlty even Firefox was updated to place its files somewhere in ~/.local for new installations (and anyone who manually moves it there).
> 
> > For example: even git stores its global configuration directly in ~/.gitconfig.
> 
> Debian does not necessarily store that file there. According to the manual, it prefers `$XDG_CONFIG_HOME/git/config` if it exists, and thats what it is doing on all my computers too.
> 
> ```
> FILES
>        By default, git config will read configuration options from multiple files:
> 
>        $(prefix)/etc/gitconfig
>            System-wide configuration file.
> 
>        $XDG_CONFIG_HOME/git/config, ~/.gitconfig
>            User-specific configuration files. When the XDG_CONFIG_HOME environment variable is not set or empty, $HOME/.config/ is used as $XDG_CONFIG_HOME.
> 
>            These are also called "global" configuration files. If both files exist, both files are read in the order given above.
> ```
> 
> > Just run ls -a under your home dir. For example, here's mine:
> 
> The issue is not discoverability, but clutter. The root of the home direcotry is for the files of the user, not random programs, and if programs put their files there users wont find their own things among all the litter. Look into your `%AppData%` folder, and imagine if all that was in your home folder on windows. It would be horrible.
> 
> Please restore the old behavior. Storing all files in `~/.local/share/sourcegit/` is not that big of a deal, but keeping and moving them to `~/.sourcegit/` is a significant downgrade. I'll probably not update until sourcegit is littering the home directory. Please understand that this is not some meaningless thing, it really benefits the users when the proper directories are used. If you need help, please let me know.

Thanks for writing a more detailed explanation then i could ever. This is what my PR is doing but was ignored. Hopefully this can be looked at again!

> @flipper
> 
> > Portable mode is for Windows only.
> 
> I dont think portable mode needs to be windows only. Why cant portable mode use the portable config directory, and then the common cache and runtime dirs?
I agree... Using XDG standard is the way to go.

PR #2129, opened 2026-02-13
fix(macOS): prevent crash on quit caused by double shutdown path

msavara:
fixes a macOS crash during app termination. App.Quit was closing MainWindow and then calling
  desktop.Shutdown(exitCode). Because the app runs with ShutdownMode.OnMainWindowClose, closing the main window already
  starts shutdown. Calling desktop.Shutdown again can re enter teardown while native AppKit and Avalonia dispatcher
  resources are already being finalized, which can end in SIGABRT on quit. The fix keeps a single shutdown path by calling
  only desktop.Shutdown(exitCode) in App.Quit, and it also adds a detailed inline comment that explains why this is
  required on macOS.

### DIFF
--- a/src/App.axaml.cs
+++ b/src/App.axaml.cs
@@ -387,7 +387,18 @@ namespace SourceGit
         {
             if (Current?.ApplicationLifetime is IClassicDesktopStyleApplicationLifetime desktop)
             {
-                desktop.MainWindow?.Close();
+                // IMPORTANT (especially on macOS):
+                // App shutdown must go through a single path only.
+                //
+                // The launcher runs with `ShutdownMode.OnMainWindowClose`, so calling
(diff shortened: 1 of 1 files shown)


love-linger:
I prefer the following solution:

```diff
 public static void Quit(int exitCode)
 {
     if (Current?.ApplicationLifetime is IClassicDesktopStyleApplicationLifetime desktop)
     {
+        desktop.ShutdownMode = ShutdownMode.OnExplicitShutdown;
         desktop.MainWindow?.Close();
         desktop.Shutdown(exitCode);
     }
     else
     {
         Environment.Exit(exitCode);
     }
 }
```

When we are calling `App.Quit` manually, we reset the `desktop.ShutdownMode` to `ShutdownMode.OnExplicitShutdown` at first.

PR #2160, opened 2026-03-03
refactor: push hotkey behavior

heartacker:
change <kbd>Ctrl+Shift+Up</kbd> to popup push, and just hit <kbd>enter</kbd>, you can push directly, easily, safely.

### DIFF
--- a/src/Views/RepositoryToolbar.axaml
+++ b/src/Views/RepositoryToolbar.axaml
@@ -106,7 +106,7 @@
     </StackPanel>
 
     <StackPanel Grid.Column="2" Orientation="Horizontal">
-      <Button Width="0" Height="0" Click="PushDirectlyByHotKey" IsVisible="False" HotKey="{OnPlatform Ctrl+Shift+Up, macOS=⌘+Shift+Up}"/>
+      <Button Width="0" Height="0" Click="PushByHotKey" IsVisible="False" HotKey="{OnPlatform Ctrl+Shift+Up, macOS=⌘+Shift+Up}"/>
       <Button Width="0" Height="0" Click="PullDirectlyByHotKey" IsVisible="False" IsEnabled="{Binding !IsBare}" HotKey="{OnPlatform Ctrl+Shift+Down, macOS=⌘+Shift+Down}"/>
       <Button Width="0" Height="0" Click="FetchDirectlyByHotKey" IsVisible="False" HotKey="{OnPlatform Ctrl+Down, macOS=⌘+Down}"/>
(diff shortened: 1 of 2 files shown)


love-linger:
We already has a `Ctrl+Shift+Up` for push directly. And `Ctrl+Up` is hardcoded in `ListBox`, `DataGrid` in `AvaloniaUI`

heartacker:
> We already has a `Ctrl+Shift+Up` for push directly. 

Yes, we know but we don't want push directly, 

Is it possible to change the action from direct push to a pop-up window? All it requires is for the user to press the Enter key.



>And `Ctrl+Up` is hardcoded in `ListBox`, `DataGrid` in `AvaloniaUI`
😢

love-linger:
> Yes, we know but we don't want push directly. Is it possible to change the action from direct push to a pop-up window? All it requires is for the user to press the Enter key.

Then this requirement is very strange. If you don't want to push directly, but pop up a panel to operate before pushing, it means that the user needs additional keyboard and mouse operations. Why not just click? Then the panel would directly appear right below the current mouse.

heartacker:
Regarding your question: "If you don't want to push directly, but pop up a panel to operate before pushing... Why not just click?"

The core motivation here is to support a fully keyboard-centric workflow for power users. Opening a panel does not necessarily mean we need to reach for the mouse. Here are the typical "no-mouse" scenarios:

Zero Mouse Usage: After writing a commit message and committing (e.g., via Ctrl+Enter), our hands are already on the keyboard. Reaching for the mouse breaks the flow.

Navigating the Panel via Keyboard: Once the hotkey opens the Push panel, a user can simply use Tab to navigate, Space to toggle options (like Force Push or Push Tags), and hit Enter to execute. The entire process requires 0 mouse clicks.

Safety Confirmation: Direct push (Ctrl+Shift+Up) is fast but sometimes risky. Many users want to trigger the popup via a hotkey just to visually double-check the target remote/branch, and then immediately press Enter to confirm. It acts as a quick safety buffer.

Since Ctrl+Up is hardcoded in Avalonia, I totally understand we can't use it. Would you be open to assigning a different default hotkey (like Alt+Up or something similar), or just adding the command so it can be triggered without the mouse?


push directly = push + enter, It is easy :)

heartacker:
I have push an new change

- change <kbd>Ctrl+Shift+Up</kbd> to popup push, and just hit <kbd>enter</kbd>, you can push directly, easily, safely.

love-linger:
now you can use command palette to do this.

PR #2176, opened 2026-03-10
fix: prevent ContainsIn flyout layout oscillation on low-perf devices

### LINKED ISSUE #2175: Commit refs tooltip sometimes has a wrong size
Image

The tooltip appears correct when it first shows, then suddenly shrinks. However, it returns to the correct size before disappearing.

OS: Deepin V25 X11

gadfly3173:
Load refs data before showing the flyout instead of loading asynchronously after it is already visible. This eliminates multiple layout passes (LoadingIcon toggle + ItemsSource update) that caused the flyout to visibly resize on slower machines.

Fixes #2175

### DIFF
--- a/src/Views/CommitBaseInfo.axaml.cs
+++ b/src/Views/CommitBaseInfo.axaml.cs
@@ -106,12 +106,16 @@ namespace SourceGit.Views
         {
             if (DataContext is ViewModels.CommitDetail detail && sender is Button button)
             {
+                var containsIn = await detail.GetRefsContainsThisCommitAsync();
+                if (containsIn is not { Count: > 0 } || !button.IsEffectivelyVisible)
+                    return;
+
                 var tracking = new CommitRelationTracking();
+                tracking.SetData(containsIn);
+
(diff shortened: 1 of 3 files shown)


love-linger:
I do not like it. The UI will get stuck when this commit is contained by too many refs. For example:

image

gadfly3173:
well, I could not find a better way to prevent the flyout resize.

love-linger:
Try to add following code to `CommitRelationTracking` to see if it solves this issue

image

gadfly3173:
image


https://github.com/user-attachments/assets/0dd2fdea-28c5-4239-89bb-0dac74c7c8d6



not work. sometimes cursor hover on the flyout would refresh it.

love-linger:
Yes, essentially, it is still because `LayoutRoot` does not correctly calculate the required size.

gadfly3173:
Closing this issue as it appears to be a timing issue within the Avalonia control itself, and there is currently no better solution.

PR #2183, opened 2026-03-12
feature: add 12-hour datetime formats

jlpatter:
I think it would be nice to have 12-hour time format options in the history so this is my attempt at adding it. Though, I'm not sure if this is the cleanest way to do it since each new time format grows the list of options exponentially. So, I'm open to suggestions if there's a better way to display the options.
Screenshot_20260311_220306


love-linger:
image

PR #2184, opened 2026-03-12

jjesus-dev:
+ Add missing strings.


love-linger:
some new strings have been added in commit 21d52c2b6b - feature: supports to switch 24-hours/12-hours (#2183)

PR #2189, opened 2026-03-13
fix: correct patch generation when discarding partial changes of staged new files

gadfly3173:
When a new file was already staged and the user tried to discard selected lines from the unstaged changes list, the generated patch was malformed due to incorrect index header format, wrong hunk line numbers, and unconditional "No newline at end of file" marker.

### DIFF
--- a/src/Models/DiffResult.cs
+++ b/src/Models/DiffResult.cs
@@ -95,53 +95,45 @@ namespace SourceGit.Models
             return rs;
         }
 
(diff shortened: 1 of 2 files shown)


love-linger:
I've pushed my fix for this bug - 1ca86ad4b8f45229d64aebe242511844adb7c1c1

gadfly3173:
index 格式、tail null 等问题可以考虑也修一下

love-linger:
- 关于 tail 为 `null` 的潜在问题已经修复
- 对于 git 对象的 hash，我倾向于不小于 8位，仅管对于空对象 `0000000` 也完全可以。

PR #1176, opened 2025-04-11
Add changed lines counter

BernatBC:
- Fixes #647 

Added and removed lines are displayed on the commit details.
image

I also plan to add these stats to files. However, I'm not sure where to place them (2 and 3)?:

image

Another PR was created a while ago (#914), but I declined it.

PR #2194, opened 2026-03-14
feature: add added/removed line counts to diff view

geviraydev:
Adds added/removed line counts to the diff view only. I considered totals for the entire change set, but didn’t want to add extra overhead, so I kept it scoped to the diff view.

This is my first PR, sorry if I missed any steps and let me know if there are any issues with this. Thanks!

image


love-linger:
#1176

PR #2206, opened 2026-03-21
Add `-dev` suffix to Version text if in DEBUG

Saibamen:
image


love-linger:
SourceGit now uses `git describe --abbrev=8 --dirty` to generate a friendly version name.

image

For example:

- `v2026.13` - app is built with the commit reference by `v2026.13` tag
- `v2026.13-2-ff872070` - app is built with 2 commits after tag `v2026.13`. `ff872070` is the HEAD while compiling.
- `v2026.13-2-ff872070-dirty` - app is built with 2 commits and some local changes after tag `v2026.13`

PR #2215, opened 2026-03-23
feature: reopen closed tab

Saibamen:
image


Saibamen:
@love-linger: Why not planned? How can we remember which tabs we accidentally closed if we have 20 tabs?

HermesSantos:
Using the recents options is not the same thing? Or am I wrong?

PR #2236, opened 2026-04-02
fix: preserve commit selection after reword

ousugo:
- keep the history list selection on the rewritten HEAD commit after using Reword
- fall back to the new current HEAD when the previously selected commit SHA no longer exists after refresh


love-linger:

PR #2260, opened 2026-04-13
feature: allow partial stage/unstage/discard for non-UTF8 text

gotoh:
### Problem

When a diff contains non-UTF8 encoded text (e.g. Shift_JIS, EUC-JP), the partial stage/unstage/discard operations produced a corrupted patch file. This happened because the diff output was read as a UTF-8 string, causing invalid bytes to be silently replaced with the Unicode replacement character (U+FFFD). The resulting patch could not be applied by `git apply`.

### Solution

- Read the raw output of `git diff` as bytes (`StandardOutput.BaseStream`) instead of a decoded string
- Store the original bytes in a new `RawContent` property on `TextDiffLine`
- Use `RawContent` when writing patch files, so the original byte sequence is preserved
- Keep the decoded `Content` string (UTF-8 with replacement chars) for display purposes only

### Files Changed

| File | Description |
|---|---|
| `src/Commands/Diff.cs` | Read diff output as `byte[]`; parse lines from raw bytes |
| `src/Models/DiffResult.cs` | Add `RawContent` to `TextDiffLine`; use raw bytes when writing patch |

### Test

1. Open a repository containing a file encoded in non-UTF8 (e.g. Shift_JIS)
2. Make a change to the file
3. In the diff view, select some lines and try **Stage Selected Lines** / **Discard Selected Lines**
4. Verify the operation succeeds and the correct lines are staged/discarded

### NOTE
This commit is a part of my non-UTF8 support.
So RawContent is also used to detect encoding and decode to text.

### DIFF
--- a/src/Commands/Diff.cs
+++ b/src/Commands/Diff.cs
@@ -45,24 +45,23 @@ namespace SourceGit.Commands
                 using var proc = new Process();
                 proc.StartInfo = CreateGitStartInfo(true);
                 proc.Start();
-
-                var text = await proc.StandardOutput.ReadToEndAsync().ConfigureAwait(false);
-
+                using var ms = new System.IO.MemoryStream();
+                await proc.StandardOutput.BaseStream.CopyToAsync(ms, CancellationToken).ConfigureAwait(false);
+                var bytes = ms.ToArray();
                 var start = 0;
-                var end = text.IndexOf('\n', start);
-                while (end > 0)
+                while (start < bytes.Length)
                 {
-                    var line = text[start..end];
-                    ParseLine(line);
-
-                    start = end + 1;
-                    end = text.IndexOf('\n', start);
+                    var end = Array.IndexOf(bytes, (byte)'\n', start);
+                    if (end < 0)
+                        end = bytes.Length;
+                    var next = end + 1;
+                    if (start <= end - 1 && bytes[end - 1] == '\r')
+                        end--;
+                    if (!_result.IsBinary)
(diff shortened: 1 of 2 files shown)


src/Commands/Diff.cs
```diff
@@ -45,24 +45,23 @@ public Diff(string repo, Models.DiffOption opt, int unified, bool ignoreWhitespa
...
-                var end = text.IndexOf('\n', start);
-                while (end > 0)
+                while (start < bytes.Length)
                 {
-                    var line = text[start..end];
-                    ParseLine(line);
-
-                    start = end + 1;
-                    end = text.IndexOf('\n', start);
+                    var end = Array.IndexOf(bytes, (byte)'\n', start);
```
love-linger, line 54: We can not use `(byte)'\n'` which actually `0x0A` to split lines without given encoding. For example:

LF in different encodings:

- UTF-8: 0x0A
- UTF-16BE: 0x000A
- UTF-16LE: 0x0A00
- UTF-32BE: 0x0000000A
- UTF-32LE: 0x0A000000

When using `0x0A` to split lines, there maybe some extra `0x00` byte remains

gotoh: UTF16/32 is not considered because `git diff` will treat it as a binary file unless we add --text. 
So sourcegit cannot handle them at all.
In the case that utf16/32 is specified as working-tree-encoding in .gitattributes, `git` will output in utf-8, so this shouldn't be a problem, I think.

PR #2275, opened 2026-04-17
fix: center diff change indicators and fix width calculation

gotoh:
## Problem

"+" indicator is not fully drawing in some fonts.

Example, with monospace font="IPAexGothic", Size=13:
image

## Changes

- Center the +/- indicators horizontally in the diff gutter
- Use Math.Max(minus.Width, plus.Width) for accurate margin width instead of only measuring the '-' character

Fixed view:
image


love-linger:
... Why not use a **TRUE** monospaced font?

gotoh:
I noticed something was off while trying out different fonts...

PR #2279, opened 2026-04-19
fix: add an missing argument of linux package's desktop entry

parintorn0:
I have added the missing argument of linux desktop entry
- Now it can open directory with an app


love-linger:
Please remove the unnecessary file `conventional_commit.json`

PR #2289, opened 2026-04-20
fix: prevent NullReferenceException in Histories async operations

gadfly3173:
Add null checks for _repo in Histories methods to prevent crashes when Repository is closed while async operations are still pending.

Bug Cause:
When Repository.Close() is called, Histories.Dispose() sets _repo to null. If an async operation (e.g., CheckoutBranchByDecoratorAsync triggered by double-clicking a branch icon) is executing concurrently, it may access the now-null _repo reference, causing NullReferenceException.

Stack Trace:

```log
System.NullReferenceException: Object reference not set to an instance of an object.
   at SourceGit.ViewModels.Repository.<CheckoutBranchAsync>d__192.MoveNext() + 0x169
--- End of stack trace from previous location ---
   at SourceGit.ViewModels.Histories.<CheckoutBranchByDecoratorAsync>d__48.MoveNext() + 0x57
--- End of stack trace from previous location ---
   at SourceGit.Views.Histories.<OnCommitListDoubleTapped>d__34.MoveNext() + 0x5a
--- End of stack trace from previous location ---
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__124_0(Object state) + 0x1a
   at Avalonia.Threading.SendOrPostCallbackDispatcherOperation.InvokeCore() + 0x19a
   at Avalonia.Threading.DispatcherOperation.Execute() + 0x4a
   at Avalonia.Threading.Dispatcher.ExecuteJob(DispatcherOperation) + 0x78
   at Avalonia.Threading.Dispatcher.ExecuteJobsCore(Boolean) + 0x387
   at Avalonia.Win32.Win32Platform.WndProc(IntPtr hWnd, UInt32 msg, IntPtr wParam, IntPtr lParam) + 0x58
   at SourceGit!<BaseAddress>+0x1166c9a

```

### DIFF
--- a/src/ViewModels/Histories.cs
+++ b/src/ViewModels/Histories.cs
@@ -143,6 +143,12 @@ namespace SourceGit.ViewModels
 
         public Models.BisectState UpdateBisectInfo()
         {
+            if (_repo == null)
+            {
+                Bisect = null;
+                return Models.BisectState.None;
+            }
+
             var test = Path.Combine(_repo.GitDir, "BISECT_START");
             if (!File.Exists(test))
             {
@@ -182,6 +188,9 @@ namespace SourceGit.ViewModels
                 return;
             }
 
+            if (_repo == null)
+                return;
+
             Task.Run(async () =>
             {
                 var c = await new Commands.QuerySingleCommit(_repo.FullPath, commitSHA)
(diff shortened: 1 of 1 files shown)


love-linger:
It's not just about `Histories`. You will get this exception when you are trying to close repository while it has an async operation running.

I believe the only solution to such problems is that when there is a action still running, the `Close` will not clear properties (since we do not have unmanaged resources), and let it recycled by system GC

love-linger:
I've pushed my fix for this bug

PR #2293, opened 2026-04-22
fix: support utf-8 passphrases in askpass

IceflowRE:
This fixes passphrases who use utf-8 characters to be correctly passed further. Based on #2288 

I have tested this change on windows.


IceflowRE:
Also added clearing the bytes and at least clearing the text field, to reduce the amount of time the passphrase is in memory.

PR #2294, opened 2026-04-23
ux: keybindings for branch rename (F2) and close Blame and FileHistories modal windows (ESC) 

asliturk:


love-linger:
I have no questions about renaming keybindings. But I do not like using `ESC` to close a non-modal dialog window (both `Blame` and `File History` are not modal dialogs)

asliturk:
I just realized they are not modal. But they feel like they are. I always feel like I should be able to close them with ESC.

There is no key combination to close these windows, would Command+F4 (Alt+F4) be more appropriate?

love-linger:
I do not like to add a keybindings to close a normal window. PS: You can already use `Alt+F4` to close `File History` or `Blame` window on Windows.

And I've pushed a commit that adds the `F2` hotkey bindings

PR #2296, opened 2026-04-23
[Feature] git init

jernejp21:
Implemented `git init` command in GUI. You can create a new git repository and also bare repository.

Icon for new repository can be changed. I couldn't find better. 
slika

If the name for the **Project Name** isn't input, the place holder text will be used for the folder name. The same is for **initial Branch Name**.
slika

For **Bare Repository**, branch name is ignored.

### DIFF
--- a/src/Commands/InitGit.cs
+++ b/src/Commands/InitGit.cs
@@ -1,0 +1,28 @@
+﻿using System.Text;
+
+namespace SourceGit.Commands
+{
+    public class InitGit : Command
+    {
+        public InitGit(string ctx, string path, string localName, string branchName, bool bareRepo)
+        {
+            Context = ctx;
+            WorkingDirectory = path;
+
+            var builder = new StringBuilder(1024);
+            builder.Append("init ");
+            
+            if (bareRepo)
+                builder.Append("--bare ");
+            else
+            { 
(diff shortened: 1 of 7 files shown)


love-linger:
You can just open/drop a folder (not git repository), `SourceGit` will open the `Initialize Repository` popup to ask if you want to run `git init` command under that folder.

If you want to clone a bare repository, you can just use `Extra Parameters`.

jernejp21:
Drag and drop doesn't work for empty folder in version 2026.09. Tested on Windows 11
On Ubuntu 24.04 deb an AppImage it doesn't work at all.
And when you open it from the menu, it won't apply group and bookmark.

slika

love-linger:
> Drag and drop doesn't work for empty folder in version 2026.09

I'm sorry. It's true. You can only trigger the `Initialize Repository` from `Open Local Repository` popup.

> And when you open it from the menu, it won't apply group and bookmark.

I've tested that `Group` will be applied to new repository (NOTE: `No Group` means it will be added to the tree root). I'll push a commit that fixes the bookmark issue

jernejp21:
Thank you for the explanation.

Too bad my PR didn't get excepted, but it was a nice programming exercise. I will close this PR.

PR #2300, opened 2026-04-24
enhance: Left/Right arrow key navigation in tree views

imurashka:
Adds Left/Right arrow key navigation in the tree view, both in Local Changes and in the Files tab of commit details. Behaves like VS Code / IntelliJ — Left collapses or walks up to the parent, Right expands or walks down to the first child.

### Local Changes

https://github.com/user-attachments/assets/43abdaa6-c680-49df-91ec-f99e3715dd31

### Files tab

https://github.com/user-attachments/assets/ac206f1b-42c1-4357-9f09-990038f406bf


love-linger:

PR #2308, opened 2026-04-27
feat: embed Git commit hash in build and display in About dialog

heartacker:
## Summary
- Embed the short git commit hash (8-char) at build time via MSBuild target, with `-dirty` suffix when the working tree has uncommitted changes
- Display the hash as `Build: {hash}` in the About dialog
- Gracefully handle missing `.git` directory or missing `git` executable (skips embedding without build failure)

## How it works
An MSBuild target runs `git describe --always --dirty=-dirty --abbrev=8 --match='___nonexistent___'` before each build and writes the result to a generated C# file compiled into the assembly as `AssemblyMetadata("GitCommitHash", ...)`. The About page reads this metadata at runtime.

## Test Plan
- [x] `dotnet build` and `dotnet publish -c Release` succeed with no warnings
- [x] About dialog displays `Build: {hash}` or `Build: {hash}-dirty`
- [x] Building without `.git` directory does not fail
- [x] Building with `.git` but without `git` installed does not fail

### DIFF
--- a/src/SourceGit.csproj
+++ b/src/SourceGit.csproj
@@ -32,6 +32,22 @@
     <DefineConstants>$(DefineConstants);DISABLE_UPDATE_DETECTION</DefineConstants>
   </PropertyGroup>
 
+  <Target Name="SetGitCommitHash" BeforeTargets="BeforeBuild"
+          Condition="Exists('$(MSBuildProjectDirectory)\..\.git')">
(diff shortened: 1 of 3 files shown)


love-linger:
It's not necessary to introduce an custom pre-built action. All .NET project has a built-in `AssemblyInformationalVersionAttribute`.

I've push my implementation for this feature

PR #2310, opened 2026-04-27
Clear notifications automatically

MarcosCobena:
I find it useful to let notifications go after a delay without having to clear them explicitly.

Backlog:
- [x] clear notifications after a delay without adding new ones
- [ ] expose this behavior at Preferences, disabled by default
- [ ] let the user customize the delay (5 s by default)

### DIFF
--- a/src/ViewModels/LauncherPage.cs
+++ b/src/ViewModels/LauncherPage.cs
@@ -132,14 +132,15 @@ namespace SourceGit.ViewModels
         {
             if (e.Action == NotifyCollectionChangedAction.Add)
             {
-                _clearNotificationsCancellationToken?.Cancel();
-                _clearNotificationsCancellationToken = new CancellationTokenSource();
                 ClearNotificationsAutomatically();
             }
         }
 
         private async void ClearNotificationsAutomatically()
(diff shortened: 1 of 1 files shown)


love-linger:
I don't want to introduce this modification.

Imagine that you open multiple repositories, perform an operation in one of them, and then switch to another to work. After a long time, you switch back to the previous one and find that the operation has been completed and there is no notification (maybe the notification has been automatically cleaned up). In this case, how to let the user know that the operation just failed?

MarcosCobena:
Thank you for the feedback.

I understand you, let me think if we can support that in any way.

I will write here down some ideas:
- moving notifications to top right corner, so do not "block" the main area and let you keep working
- hiding notifications after a delay, but user can still open and close them as before

love-linger:
The reason why the notifications(errors) are displayed in the middle (and just under the popup) of the window is:

Error messages/notifications are usually displayed after an action complete/failed. At that time, to clear notifications, user can just move mouse down and click.

MarcosCobena:
@love-linger thank you again; I am better moving this into an issue.

Issue #2318, opened 2026-05-04
`enable_thinking` parameter sent to OpenAI-compatible APIs that don't support it


Harsche:
When using an OpenAI-compatible provider (Groq in my case), the AI commit message generation fails with HTTP 400 because SourceGit is sending `enable_thinking` in the request body, which Groq does not support.

<img width="515" height="398" alt="Image" src="https://github.com/user-attachments/assets/7364d962-0d31-4d24-8c68-3d99b9f939c6" />

dan2dev:
Same issue here.

```
Asking AI to generate commit message...


[ERROR]
HTTP 404 (invalid_request_error: )
Parameter: model

This is not a chat model and thus not supported in the v1/chat/completions endpoint. Did you mean to use v1/completions?
```

dougcunha:
Same here

<img width="735" height="489" alt="Image" src="https://github.com/user-attachments/assets/fd1af396-c963-4366-8f32-d3092713702e" />

love-linger:
Fixed. Instead of directly disabling `thinking mode` (the way to disable `thinking mode` varies across different services, and some services do not allow unsupported parameters to be passed), the current approach is to return the `reasoning_content` to the AI service as-is if it is included in the response.

love-linger:
You can download the latest CI build from GitHub Action and test it.

dougcunha:
Thank you @love-linger . It's working now.

PR #2319, opened 2026-05-05
fix: Add IsAnthropicCompatible property and disable thinking in Agent

dan2dev:
Add IsAnthropicCompatible property and disable thinking in Agent when is not Anthropic. #2318


love-linger:
I've pushed my fix for this BUG

PR #2321, opened 2026-05-05
feature: allow copying ref names from commit info panel

ivnnv:
Adds a small copy button next to each ref pill in the commit info panel. Clicking it copies the ref name (e.g. `main`, `origin/feature`) to the clipboard, matching the existing copy affordance on the SHA row.

Screenshot 2026-05-05 at 14 02 30

The Histories grid is intentionally untouched (copy buttons there would clutter the dense rows). One side effect to flag: wrap in the REFS row now happens between `[pill+button]` groups instead of inside one continuous presenter, which is the natural behavior for an `ItemsControl` + `WrapPanel` layout.

## Test plan

- [ ] Select a commit with refs and click a copy icon in the INFORMATION panel's REFS row — the ref name lands in the clipboard.


love-linger:

ivnnv:
you are welcome :)

PR #2322, opened 2026-05-05
feature: add Compare with default base on current branch menu

ivnnv:
Other local branches in the sidebar already have a one-click "Compare with HEAD" entry, but the current branch's right-click only offers the free-form "Compare with..." palette. Adds a "Compare with `<default>`" entry that opens the existing Compare window pre-filled with the repo's default base branch.

Screenshot 2026-05-05 at 17 51 04

The default branch is resolved from `refs/remotes/<remote>/HEAD` (the same source GitHub/GitLab use), with the remote chosen via the same fallback order used elsewhere in the app: current branch's upstream remote → `Settings.DefaultRemote` → first remote. The entry is hidden when the remote HEAD is unset or when the resolved base equals the current branch (so it never shows "Compare master with master").

## Test plan

- [ ] Right-click the current branch and confirm "Compare with `<default>`" appears, opens the existing Compare window pre-filled.

### DIFF
--- a/src/Commands/QueryDefaultBranch.cs
+++ b/src/Commands/QueryDefaultBranch.cs
@@ -1,0 +1,27 @@
+using System.Threading.Tasks;
+
+namespace SourceGit.Commands
+{
+    public class QueryDefaultBranch : Command
+    {
+        public QueryDefaultBranch(string repo, string remote)
+        {
+            WorkingDirectory = repo;
+            Context = repo;
+            Args = $"symbolic-ref --short refs/remotes/{remote}/HEAD";
+            RaiseError = false;
+        }
+
+        public string GetResult()
+        {
+            var rs = ReadToEnd();
+            return rs.IsSuccess ? rs.StdOut.Trim() : string.Empty;
+        }
+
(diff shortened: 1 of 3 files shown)


love-linger:
`Compare with HEAD` is designed to allow quick comparison with the current branch (excluding uncommitted files - that's why we use `HEAD` instead of current branch name), and it has nothing to do with the default(main) branch. 

I don't think `SourceGit` needs to integrate the feature of comparing with `<remote>/HEAD`:

- When fallback to `<remote>/HEAD`, it means that the `<remote>/HEAD` is not the upstream of current branch. Therefore, I do not think we need to provide an individual context menu entry for this purpose. If users need it, they can select any branch to be compared from the `Compare with...` context menu entry.
- User may confuse about that why non-current local branches have a context menu to compare with current branch's HEAD but the current local branch has a context menu to compare with remote's HEAD?

So, I pushed another commit that adds a quick context menu entry for all local branches to compare with its upstream

ivnnv:
ok np, at least it inspired a different take on the problem. ill keep my version in my fork since it covers the no-upstream case too. thanks for reviewing

Issue #2245, opened 2026-04-07
macOS New branch button is gone


nilsnilsnils:
Latest Version . New branch button is gone. Can't create  new branch !

love-linger:
It's here:

<img width="566" height="460" alt="Image" src="https://github.com/user-attachments/assets/819859ff-fe7a-4dd7-9171-03e8c996b382" />

Just like the `Add Remote`, `New Tag`, `Add Submodule` and `Add Worktree` buttons.

And you can also use `Ctrl+B` hotkey to open `Create Branch` popup.

love-linger:
I'm sorry. The hotkey is `Command+B` on macOS.

And you can also create a new branch in follow ways:

- Context menu entry `Create Branch...` of selected branch
- Context menu entry `Create Branch...` of selected tag
- Context menu entry `Create Branch...` of selected commit

nilsnilsnils:
Okay . But why remove the button which is used so often ? Can I configure the toolbar ?

love-linger:
As the above image shows, makes the UI/UX design more reasonable.

Users manage theirs local branches in LOCAL BRANCHES tree, remotes in REMOTES tree, .etc. And keeps the toolbar buttons only care about repository itself.

love-linger:
Forget to say, that you can also create a new branch from `Command Palette` (Ctrl+Shift+P/⌘+⇧+P)

<img width="556" height="577" alt="Image" src="https://github.com/user-attachments/assets/580d00ea-f1ed-40cb-a574-0678ac44f3b0" />

PR #2346, opened 2026-05-15
enhance: worktree and branch UI improvements, rename remote branch, remove tracking reference

briepala:
### Branch Operations

**Rename remote branch**: New "Rename..." context menu on remote branches opens a popup that pushes the current commit to a new branch name on the remote, deletes the old one, and updates the tracking of any local branches that tracked the old remote branch.

**Remove tracking reference**: New "Remove Tracking Reference" context menu item on local branches that have an upstream set. Calls `git branch --unset-upstream` directly.

**Rename local branch with remote**: Checkbox "Also rename remote branch" in the rename dialog. Fixes:
- Correctly deletes the actual remote branch name (not the local branch name) when they differ
- Pushes the tracking remote's HEAD commit rather than the local ref
- Updates the local branch's upstream tracking to the new remote branch
- Checkbox is disabled when local and remote branches are at different commits (`CanRenameRemote` guard)

**Delete remote branch**: When deleting a remote branch directly, all local branches that were tracking it get their upstream unset automatically.

**Delete branch with worktree**: Confirmation checkbox to force-remove the worktree before deleting the branch (avoids git's rejection of deleting a branch checked out in a worktree).

### Toolbar
- Repository folder name shown in bold to the left of the current branch name.
- New Branch and Worktree toolbar buttons for quick access to create-branch and add-worktree dialogs.
- All 7 action buttons (Fetch, Pull, Push, Stash, Apply, Branch, Worktree) display text labels beneath their icons.

### Localization
- New resource strings added to en_US and all translated locale files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ ehthumbs_vista.db
 
 bin/
 obj/
+publish/
 # ignore ci node files
 node_modules/
 package.json
--- a/src/Resources/Locales/de_DE.axaml
+++ b/src/Resources/Locales/de_DE.axaml
@@ -324,6 +324,7 @@ $1, $2, … Werte der Eingabe-Steuerelemente</x:String>
   <x:String x:Key="Text.DeleteBranch.Branch" xml:space="preserve">Branch:</x:String>
   <x:String x:Key="Text.DeleteBranch.IsRemoteTip" xml:space="preserve">Du bist dabei, einen Remote-Branch zu löschen!!!</x:String>
   <x:String x:Key="Text.DeleteBranch.WithTrackingRemote" xml:space="preserve">Auch Remote-Branch ${0}$ löschen</x:String>
+  <x:String x:Key="Text.DeleteBranch.WithWorktree" xml:space="preserve">Auch Worktree unter ${0}$ entfernen</x:String>
   <x:String x:Key="Text.DeleteMultiBranch" xml:space="preserve">Mehrere Branches löschen</x:String>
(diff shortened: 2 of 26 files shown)


love-linger:
First of all, keep your PR focus on a single issue.

> Remove tracking reference

It's not necessary. You can change the upstream from `Set Tracking Branch...` even if you want to unset it

> Rename remote branch

There's no plan to add this feature to `SourceGit`. If you want, I recommand you to add a custom action with a user script to do this. 

I want all the `SourceGit` users to know that we can not rename a remote branch. We can just push a new one and delete the old one (if it is needed). At the same time, in some collaborative projects, deleting a remote branch may require additional operations. For example, it may be necessary to lock the remote branch first to ensure that no new commits are pushed to this remote branch during the deletion process.

> Delete branch with worktree

I'd like to let user's know why this branch can not be deleted directly.

> Toolbar - Repository folder name shown in bold to the left of the current branch name.

Not planed. If you have more than one repositories that have the same name, you can rename it in `SourceGit` (using context menu of repository tab) which will not change the folder name but give a friendly name displayed in tabs

> Toolbar - New Branch and Worktree toolbar buttons for quick access to create-branch and add-worktree dialogs.

Not planned. In fact, I just removed the `New Branch` button from toolbar since `v2026.07`. See #2245

> Toolbar - All 7 action buttons (Fetch, Pull, Push, Stash, Apply, Branch, Worktree) display text labels beneath their icons.

Not planned. I've removed the button labels since `v4.9` (2021/08/13)

PR #2365, opened 2026-05-20
ux: Add config option to show containing refs directly in the commit detail view

floriandammeyer:
Adds a new config option to show the containing refs directly in the commit detail view. The default is `false` so that the current behavior of the app stays the default.

When enabled, the tooltip "commit is contained by" is hidden and instead a new row "CONTAINED BY" is shown above the "REFS" in the commit detail view. 

The containing refs are loaded asynchronously to not block the app.

New option in the app settings (red border is only to highlight what changed, not part of the app):
grafik

Updated commit details view:
grafik

### DIFF
--- a/src/Converters/DoubleConverters.cs
+++ b/src/Converters/DoubleConverters.cs
@@ -19,5 +19,8 @@ namespace SourceGit.Converters
 
         public static readonly FuncValueConverter<double, Thickness> ToLeftMargin =
             new FuncValueConverter<double, Thickness>(v => new Thickness(v, 0, 0, 0));
+
+        public static readonly FuncValueConverter<bool, double> ToContainingRefsMaxHeight =
+            new FuncValueConverter<bool, double>(v => v ? double.PositiveInfinity : 24.0);
     }
 }
--- a/src/Models/RepositorySettings.cs
+++ b/src/Models/RepositorySettings.cs
@@ -36,6 +36,18 @@ namespace SourceGit.Models
             set;
         } = false;
 
+        public int ShowContainingRefsInCommitDetail
+        {
+            get;
+            set;
+        } = 0;
+
+        public int ContainingRefsDefaultExpansion
+        {
+            get;
+            set;
+        } = 0;
+
         public string PreferredOpenAIService
         {
             get;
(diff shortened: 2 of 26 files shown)


love-linger:
I do not like this idea.

- It's not a primary (important) info of a commit.
- It is displayed upon the `REFS` (which I think is more important than it). This has caused confusion of information to a certain extent.
- `SourceGit` already provides a way to view the refs that contains selected commit.

image

What you can see in above screenshot is: when selects a commit that is contained by lots of refs, the whole area is used by `CONTAINED BY`. And this issue will affected all its parents. User can not see the REFS/MESSAGE/CHANGES (which I think is more important) without scrolling it.

I'd like to keep current design.

floriandammeyer:
I see your point and have incorporated your remarks into the PR:

- switched order of the "REFS" and "CONTAINED BY" rows so that the direct refs are above the containing refs and so that the order of "parents", "children" and "refs" is left exactly the way it is now
- the row "contained by" is now collapsed into a single row by default if there are too many refs. The row can be expanded using a chevron icon if needed
- the feature is entirely configurable globally and on a per-repository basis: if app settings are not changed, the current behavior (using the tooltip for the containing refs) is used and completely unchanged. if enabled in the settings, the new row "containing refs" is being shown. by default this row is shown in a collapsed state. if the according option in the setting is changed, this can also default to the expanded state. this way this feature does not change anything for existing users which are happy with the tooltip. if a user wants to change the display only for a specific repository (for example because there is one repository with a lot of refs), it can also be overwritten in the repository settings in a new "General" tab.

I understand that it may not be helpful to all users, which is why I made the feature completely configurable and optional. Without changing the default settings, the app still behaves exactly as it did before. But for me, and maybe for others too, seeing the containing refs immediately can be helpful, so I ask you to think about it again.

## Screenshots to outline the changes

Added configuration options in the app settings
Bildschirmfoto 2026-05-21 um 11 40 14

Possibility to overwrite the app settings per repository
Bildschirmfoto 2026-05-21 um 11 46 25

With default settings, this feature is disabled so the commit detail view looks exactly the same as it does now (= Tooltip icon and no "contained by" row):
Bildschirmfoto 2026-05-21 um 11 57 07

Collapsed state of the containing refs row with chevron icon:
Bildschirmfoto 2026-05-21 um 11 41 12

Expanded state of the containing refs row with chevron icon:
Bildschirmfoto 2026-05-21 um 11 44 03

Refs row is above containing refs row to keep the current order in the table:
Bildschirmfoto 2026-05-21 um 11 46 57

TomW-Skyline:
> * `SourceGit` already provides a way to view the refs that contains selected commit.

@love-linger what is currently the easiest way to see this list in SourceGit?

love-linger:
Click the second button next to the commits SHA in commit details panel

PR #2411, opened 2026-06-01
feature: New option to ignore casing changes, ignore casing and ignore whitespace button also for unstaged changes file list and commit details

Hurricane31337:
I implemented a new option to ignore casing changes. And added this button and the one to ignore whitespace in the unstaged changes file list and the commit changes in the commit details in the commit history. This gives you more flexibility when checking very large merges for example, when some files only have minor spacing or casing changes (depending on your programming language, the casing can be irrelevant. These features help us to save time in our daily workflow, maybe you have a use for it, too. I also added the translation to all supported languages. I used Claude Opus 4.8 and didn't do everything by hand.

## Ignore casing:
image
image

## Improved "ignore whitespace" to also ignore added/removed empty lines:
image
image

## New option to hide files with only whitespace or casing changes in changes view:
image
image

## New option to hide files with only whitespace or casing changes in commit details:
image
image

### DIFF
--- a/src/Commands/Diff.cs
+++ b/src/Commands/Diff.cs
@@ -20,9 +20,10 @@ namespace SourceGit.Commands
         private const string PREFIX_LFS_DEL = "-version https://git-lfs.github.com/spec/";
         private const string PREFIX_LFS_MODIFY = " version https://git-lfs.github.com/spec/";
 
-        public Diff(string repo, Models.DiffOption opt, int unified, bool ignoreWhitespace)
+        public Diff(string repo, Models.DiffOption opt, int unified, bool ignoreWhitespace, bool ignoreCase = false)
         {
             _result.TextDiff = new Models.TextDiff();
+            _ignoreCase = ignoreCase;
 
             WorkingDirectory = repo;
             Context = repo;
@@ -32,7 +33,7 @@ namespace SourceGit.Commands
             if (Models.DiffOption.IgnoreCRAtEOL)
                 builder.Append("--ignore-cr-at-eol ");
             if (ignoreWhitespace)
-                builder.Append("--ignore-space-change ");
+                builder.Append("--ignore-all-space --ignore-blank-lines ");
             builder.Append("--unified=").Append(unified).Append(' ');
             builder.Append(opt.ToString());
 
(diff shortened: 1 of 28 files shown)


goran-w:
NOTE: Recently merged PR #2436 moves the ignore-whitespace button to the right (to avoid it jumping around when toggled, if the Diff is all-whitespace).

love-linger:
I don't like this pull request.

- Running `git diff` for every change under `LOCAL CHANGES` will definitely slow down the list loading, especially when there are a large number of files. 
- It is not recommended to hide any files other than untracked files in `LOCAL CHANGES` page, as this can easily lead to commit errors. That's why `SourceGit` prompts users for confirmation before committing when files in `STAGED` are hidden due to a searching operation.
- There are no plans to add the `Ignore Case` feature.

goran-w:
> * Running `git diff` for every change under `LOCAL CHANGES` will definitely slow down the list loading, especially when there are a large number of files.

That's definitely an issue! 😅

> * It is not recommended to hide any files other than untracked files in `LOCAL CHANGES` page, as this can easily lead to commit errors. That's why `SourceGit` prompts users for confirmation before committing when files in `STAGED` are hidden due to a searching operation.

To be fair, I think the suggested feature could be viewed as yet another way to filter the changes, similar to the search-filtering that's already there. The same prompt for confirmation before committing when filters are active would apply here too.

However, these filters should ideally work on already available info about the type/name/path of the item-change, and not on the contents of the Diff itself, which would rule out the proposed filters - unless there's some Git option which can be used to VERY QUICKLY determine files which yield no diff-lines given a set of specified ignore-options.

Leaving these `Local Changes` filters aside, the new/modified Diff-filters could be viewed as a separate suggestion.
* I'm a bit hesitant about the usefulness of then suggested new "Ignore Case Changes" feature. It could hide real mistakes, where unintended case changes are missed among the intended ones. Also, how often do you keep toggling the casing in text - and wouldn't you want to review these changes?
* The suggested modification to additionall "ignore added/removed empty lines" could be a useful addition. However:
  * Please don't replace `--ignore-space-change` with `--ignore-all-space`, since the latter option would hide actual 'semantic' differences! (For example, it would treat "be at" and "beat" as identical...)
  * IMHO, the `--ignore-blank-lines` option could well be added (as suggested) into the current implementation of the ignore-whitespace toggle (in combination with the already present `--ignore-space-change` option).
  * It should be noted though, that the `--ignore-blank-lines` option in Git has a somewhat counter-intuitive implementation. I would expect it to ignore any added/removed blank lines, and it does that - but ONLY when there are NO OTHER unignored changes in the same file... 🤔 (Still, it would be a useful addition wherever it applies, since it would help in quickly skipping over files that have only empty-line changes.)

Hurricane31337:
If it's not beneficial for everyone in your opinion, feel free to just close the PR.
I just did it because we had a HUGE merge and a large part of it were just spacing and casing changes that were irrelevant in our case (VB.NET, some auto-renaming from a Visual Studio inconsistency). I'm aware that in general, you should be able to see what you're committing, but in my/our special case it was a time saver.

Mingun:
Instead of introducing Diff filters in case-by-case it would be better to add ability to normalize line before compare, like in KDiff3 and ability to enable those filters in diff view. But that will require implement the diff algorithms in the SourceGit, while currently it just parses the `git` output.

Issue #2435, opened 2026-06-12
[Bug] When Diff is all-whitespace, toggling 'Ignore Whitespace Changes' Off won't work


goran-w:
Currently, when a diff contains only whitespace changes, then `Ignore Whitespace Changes` can only be toggled On - trying to toggle it Off will do nothing!

The reason is that when On, the `Content` is no longer a (descendant of) class `TextDiffContent` but the (standalone) class `NoOrEOLChange`. To fix this, the if-condition could be moved down to cover just the last part, where the `TextDiffContent ctx` is actually used/required...

<img width="793" height="375" alt="Image" src="https://github.com/user-attachments/assets/1531c18d-6fe2-4ba2-a7fc-4430642416ad" />

On a much related note, the toggle button for `Ignore Whitespace Changes` needs to be moved "two steps" to the right, so that it doesn't move sideways when toggling it (in the case of an all-whitespace Diff).

When Off, it's the 4th button from the right:
<img width="363" height="81" alt="Image" src="https://github.com/user-attachments/assets/5b979a23-5642-4796-b007-6a48edb0d0bc" /> 

When On, it's the 2nd button from the right:
<img width="334" height="75" alt="Image" src="https://github.com/user-attachments/assets/18673b63-10c8-44a7-b797-5b5e3a785725" />

Having a toggle-button "jump around" when toggling it is bad design and very annoying - the user should be able to keep toggling the button On and Off (without moving the mouse pointer) to see the effect it has. (In this case, the second toggle-click could end up on the `Side-by-Side Diff` toggle button which takes its place.)

love-linger:
Merged

PR #2436, opened 2026-06-12
Fix issue where ignore-whitespace can't be toggled (#2435)

goran-w:
Also move the toggle-button to avoid having it jumping around when toggled (which is a UI-design no-no).

This fixes issue #2435.

(Additionally, renamed nearby variable from _entireFileLine to _entireFileLines, for clarity and consistency with variable _unifiedLines.)

### DIFF
--- a/src/ViewModels/DiffContext.cs
+++ b/src/ViewModels/DiffContext.cs
@@ -85,12 +85,12 @@ namespace SourceGit.ViewModels
 
         public void CheckSettings()
         {
+            var pref = Preferences.Instance;
+
             if (Content is TextDiffContext ctx)
             {
-                var pref = Preferences.Instance;
-
-                if ((pref.UseFullTextDiff && _info.UnifiedLines != _entireFileLine) ||
-                    (!pref.UseFullTextDiff && _info.UnifiedLines == _entireFileLine) ||
+                if ((pref.UseFullTextDiff && _info.UnifiedLines != _entireFileLines) ||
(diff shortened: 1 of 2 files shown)


love-linger:
- Please run `dotnet format src/SourceGit.csproj` first
- I'd like to change the `CheckSettings` to

```csharp
var pref = Preferences.Instance;

if (Content is TextDiffContext ctx)
{
    if ((pref.UseFullTextDiff && _info.UnifiedLines != _entireFileLine) ||
        (!pref.UseFullTextDiff && _info.UnifiedLines == _entireFileLine) ||
        (pref.IgnoreWhitespaceChangesInDiff != _info.IgnoreWhitespace))
    {
        LoadContent();
        return;
    }

    if (ctx.IsSideBySide() != pref.UseSideBySideDiff)
        Content = ctx.SwitchMode();
}
else if (Content is Models.NoOrEOLChange)
{
    if (pref.IgnoreWhitespaceChangesInDiff != _info.IgnoreWhitespace)
        LoadContent();
}
```

goran-w:
> * Please run `dotnet format src/SourceGit.csproj` first

Fixed. But could we have this format step run automatically when building in Visual Studio? It's annoying when a PR check fails because of a missed step that could've been automated...

> * I'd like to change the `CheckSettings` to

Done.

love-linger:
> But could we have this format step run automatically when building in Visual Studio? It's annoying when a PR check fails because of a missed step that could've been automated

I'm not sure if we add some pre-build task will work for all IDEs. Instead, before I push something to this repository, I'll run a custom action to do this job.

image

PR #2440, opened 2026-06-15
fix: remote expanded nodes lost after checkout or create/rename branch

JC-Chung:
`RefreshAfterCreateBranch`, `RefreshAfterCheckoutBranch`, and
`RefreshAfterRenameBranch` call `BuildBranchTree` with an empty remotes
list, causing all `refs/remotes/*` entries in `ExpandedBranchNodesInSideBar`
to be incorrectly treated as invalid and removed.

Fix by passing `_remotes` instead of `[]`.

### DIFF
--- a/src/ViewModels/Repository.cs
+++ b/src/ViewModels/Repository.cs
@@ -806,7 +806,7 @@ namespace SourceGit.ViewModels
                     locals.Add(b);
             }
 
-            var builder = BuildBranchTree(locals, []);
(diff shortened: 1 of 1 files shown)


love-linger:
This PR only fixes the issue for the remote node itself. When some remote branches has folders, the states of them are still missing.

I've pushed my fixes for this issue. See commit c6b942e497e7fe009cba6547154103c31ed4987f

PR #2445, opened 2026-06-15
feature: author filter and go-to-SHA jump in commit history

victor-ca:
## What

Adds two affordances to the commit history column headers:

- **Author filter** — a funnel icon in the Author column header opens a filter popup (search box + contributor picker) that hides non-matching commits and rebuilds the graph. Filtering is debounced and leaves the full commit set intact.
- **Go-to-SHA** — a search icon in the SHA column header jumps/scrolls to the commit matching a typed prefix.

## Screenshots
image
image

### DIFF
--- a/TRANSLATION.md
+++ b/TRANSLATION.md
@@ -6,7 +6,7 @@ This document shows the translation status of each locale file in the repository
 
 ### ![en_US](https://img.shields.io/badge/en__US-%E2%88%9A-brightgreen)
 
-### ![de__DE](https://img.shields.io/badge/de__DE-91.08%25-yellow)
+### ![de__DE](https://img.shields.io/badge/de__DE-90.63%25-yellow)
 
 <details>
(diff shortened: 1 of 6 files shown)


love-linger:
Please use `Search Commit` instead

image

victor-ca:
> Please use `Search Commit` instead
> 
> image

I do use Search Commit - this is meant to complement it, not replace it.

The main difference: Search Commit needs me to type the author's name, and I don't always remember it. Filtering straight from the author column lets me pick someone I recognize without knowing the exact name up front. The in-graph access is a secondary convenience on top of that.

Happy to drop it if you'd still rather keep everything in Search Commit - just wanted to explain the use case first.

love-linger:
I've pushed two commit that:

- Show suggestions when searching commits by author or committer
- Directly select the result when searching commit by SHA

PR #2455, opened 2026-06-17
[enhancement] Add flag '--ignore-blank-lines' when ignoring whitespace

goran-w:
Inspired by a discussion in PR #2411.
* With the added option `--ignore-blank-lines`, some more cases of changes in (added / removed) empty lines will be ignored (but apparently not when there are also other unignored changes in the same file).
* Still, it's a useful addition wherever it applies, since it helps in quickly skipping over files that have only empty-line / line-break changes (by displaying the `No changes or only EOL changes` panel).
* Also renamed argument `unified` to `numContextLines`, for clarity.


goran-w:
@love-linger NOTE: This PR incurs a very minor change, and I see no adverse effects of adding the additional Git flag `--ignore-blank-lines` when we're already applying `--ignore-space-change`.

PR #2461, opened 2026-06-18
feature: locale-aware commit messages

YuHaoLiaoTony:
The "Generate commit message with AI" feature previously sent a fully English prompt to
the AI, so commit messages were always generated in English — regardless of the user's
UI language setting.

This PR adds a single language instruction line to the prompt, derived from the app's
current locale. For example, when the UI is set to Traditional Chinese, the prompt now
includes "Write the commit message in Traditional Chinese."

- `Agent.cs`: added optional `locale` parameter to `GenerateCommitMessageAsync`,
  injects language instruction using the existing `DiffPrompts.GetOutputLanguage()` mapping
  (supports 16 languages)
- `AIAssistant.cs`: passes `Preferences.Instance.Locale` to the agent call
- Backward compatible — defaults to English when locale is not provided
- Works with existing `AdditionalPrompt` customization (language instruction is placed
  before `AdditionalPrompt`)

2 files changed, 3 insertions, 2 deletions.

### DIFF
--- a/src/AI/Agent.cs
+++ b/src/AI/Agent.cs
@@ -14,7 +14,7 @@ namespace SourceGit.AI
             _service = service;
         }
 
-        public async Task GenerateCommitMessageAsync(string repo, string changeList, Action<string> onUpdate, CancellationToken cancellation)
+        public async Task GenerateCommitMessageAsync(string repo, string changeList, Action<string> onUpdate, CancellationToken cancellation, string locale = null)
         {
             var chatClient = _service.GetChatClient();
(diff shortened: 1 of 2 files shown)


love-linger:
You can just add a single line in `Preferences > AI > Additional Prompt`

```
- Response in Traditional Chinese
```

In this way, you can specify all the human language instead of thoes supported by `SourceGit`

YuHaoLiaoTony:
Agreed, that approach is more flexible. I'll close this. Thanks for the review.

PR #2465, opened 2026-06-20
ux: increase mouse scroll speed for launcher tabs

b-w:
When having a lot of repository tabs open, scrolling through them becomes a chore because the scroll speed is so slow.

This change improves UX by simply increasing the speed.

Note: I would have preferred to do this by configuring the control itself, but Avalonia doesn't expose any options for it, hence the slightly clunky for loop.

### DIFF
--- a/src/Views/LauncherTabBar.axaml
+++ b/src/Views/LauncherTabBar.axaml
@@ -15,7 +15,7 @@
                   Classes="icon_button"
                   Width="18" Height="28"
                   Margin="0,2,0,0"
-                  Click="ScrollTabsLeft"
+                  Click="OnScrollTabsLeft"
                   IsVisible="{Binding #ThisControl.IsScrollerVisible}">
       <Path Width="8" Height="14" Stretch="Fill" Data="{StaticResource Icons.TriangleLeft}"/>
     </RepeatButton>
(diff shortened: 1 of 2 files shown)


love-linger:
I've pushed my implementation for this feature. See f56ec6c6cc6b351b75357fa2a895f546d77920c4

b-w:
@love-linger in your implementation, mouse scrolling doesn't work at all anymore; you need to hold shift. not sure what the reasoning behind it is, but that's a ux regression imo.

love-linger:
In `SourceGit`, when you want to horizontal scrolling some page, you must hold the `SHIFT`. Otherwise, it means to scroll that page vertically.

b-w:
in the tab bar, vertical scrolling does not exist. you can ONLY do horizontal scrolling. having a modifier key for that makes no sense, and will force users to always hold an extra button for no reason just to use the only possible scrolling option.

i've never seen a tab bar in an app work like this, and again I think this will be an annoying ux regression.

love-linger:
Using `SHIFT` is the only way to distinguish between scrolling with a mouse wheel and two-finger scrolling on a trackpad.

In `Avalonia`, the `e.Pointer.Type` also returns `Mouse` when scrolling via a trackpad.

love-linger:
I've pushed another commit.

0721743e45 - ux: convert vertical scrolling to horizontal scrolling for launcher tabs

b-w:
had a look and it seems to work quite well, thanks for your efforts!

I guess this PR is no longer needed so I'll close it.

PR #2474, opened 2026-06-23
[fix + enhancement] Improve handling of filemode changes

goran-w:
This PR fixes some issues in the current handling of Git diffs containing files with a 'type changed' (T) state (i.e changed filemode). Such files were sometimes completely missing from the list of changes in SourceGit!

It also makes 'filemodes' displayed in the UI more readable, by showing the name of the mode instead of its 'magic number'.

image

image

Detailed changes:
* Add handling of ChangeState.TypeChanged 'T' in `CompareRevisions` and `QueryFileHistory` (it's already handled in `QueryLocalChanges` and `QueryStagedChangesWithAmend`).
* `DiffResult` class:
  * Change type of `OldMode` and `NewMode` from `string` to `int`, for more convenient handling.
  * Make the `FileModeChange` property return a (human-readable, localized) name instead of a 'magic number'.
  * NOTE: The 6 new localized strings (`Text.FileMode.*`) are only provided for `en_US`, so far.
* `Diff` class:
  * Refactor the `ParseLine()` method by delegating to 3 new methods (for parsing of diff-header, chunk-start and chunk-body).
  * Add new private bool-variable `_isInChunk`, to keep track of when we're parsing the 'inside' of a chunk-body.
  * Add new enum `Indicator` to be used instead of repeated 'raw' char constants, which also makes the code easier to follow.
  * Improve the parsing within `ParseLine()` and related methods to correctly handle 'type changed' diffs, which actually contain two full diffs in one. (I.e 1st diff-header, 1st chunk, 2nd diff-header, 2nd chunk - see the 2nd screenshot above for an example!)
  * Make `ParseFileModeChange()` parse file-mode strings into integers (they are actually 'octal' numbers, but we can treat them as 'decimal' numbers since we only ever check for equality against known values).
  * Refactor `ParseLFSChange()` to make better use of the chunk-body indicators (i.e the first `char` in each chunk-body line).

### DIFF
--- a/src/Commands/Diff.cs
+++ b/src/Commands/Diff.cs
@@ -16,9 +16,18 @@ namespace SourceGit.Commands
         [GeneratedRegex(@"^index\s([0-9a-f]{6,64})\.\.([0-9a-f]{6,64})(\s[1-9]{6})?")]
         private static partial Regex REG_HASH_CHANGE();
 
-        private const string PREFIX_LFS_NEW = "+version https://git-lfs.github.com/spec/";
-        private const string PREFIX_LFS_DEL = "-version https://git-lfs.github.com/spec/";
-        private const string PREFIX_LFS_MODIFY = " version https://git-lfs.github.com/spec/";
+        private const string LFS_SPECIFIER = "version https://git-lfs.github.com/spec/";
+        private const string LFS_OID_PREFIX = "oid sha256:";
+        private const string LFS_SIZE_PREFIX = "size ";
+
+        private enum Indicator
+        {
+            ChunkHeader = '@',
+            Context = ' ',
+            Old = '-',
+            New = '+',
+            Special = '\\',
+        }
 
         public Diff(string repo, Models.DiffOption opt, int numContextLines, bool ignoreWhitespace, bool ignoreCRAtEOL)
         {
@@ -81,7 +90,12 @@ namespace SourceGit.Commands
             }
             else
             {
-                ProcessInlineHighlights();
+                if (_isInChunk)
+                {
+                    ProcessInlineHighlights();
+                    _isInChunk = false;
+                }
+
                 _result.TextDiff.MaxLineNumber = Math.Max(_newLine, _oldLine);
                 _result.TextDiff.OldMode = _result.OldMode;
                 _result.TextDiff.NewMode = _result.NewMode;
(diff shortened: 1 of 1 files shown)


gadfly3173:
> Make the FileModeChange property return a (human-readable, localized) name instead of a 'magic number'.

I don't think this is a widely-used display format. It could be confusing for users who are already familiar with conventional file mode notations. Moreover, given the limited space in this UI region, supporting i18n for languages such as Russian , Portuguese and etc. would introduce considerable translation overhead — the localized file mode string might even end up being longer than the actual file path.

Fork:
image

Gitk:
image

AI translation example:
image

goran-w:
@gadfly3173 wrote:
> I don't think this is a widely-used display format. It could be confusing for users who are already familiar with conventional file mode notations. Moreover, given the limited space in this UI region, supporting i18n for languages such as Russian , Portuguese and etc. would introduce considerable translation overhead — the localized file mode string might even end up being longer than the actual file path.

I see your point, but far from every user is familiar with these 'magic numbers', or have them memorized even after prolonged usage. (And for my own part, I very much prefer the "friendlier" names!)

As for localized name length, the longest name seems to be "Symbolic Link" which can be shortened as "Symlink" / "Link" or similar, and the other longer names can also me shortened (by dropping "File", for example). Let me add a commit for this...

I'd like to keep the friendly names displayed - but I'll extend the existing tooltip "File Mode Changed" with the numbers. Would that solve your issue?

(If not, I would add a Preference setting to display the numbers again - but friendly names should be the default, since they are much more helpful for new users!)

goran-w:
@gadfly3173 The longer names are now shortened (`Normal`, `Executable`, `Symlink`), and the numbers are now shown in tooltip:

image

love-linger:
I do not like this PR

- I'd like to keep current design and show translated file type change in tooltip instead
- I do not like the changes about `Commands.Diff`. For example:

I do not think the followin change is much more faster or easier to understand

image

You still need to compare the last string after checking the first character.

love-linger:
I'll try to implement this feature

goran-w:
@love-linger Why won't you let me fix my own PR instead. 😢

goran-w:
@love-linger wrote:
> I do not think the followin change is much more faster or easier to understand

It's not faster, but it's more consistent and easier to follow since it uses the Indicator constants. (But I see now that we should add constants for the oid and size strings, too.)

love-linger:
Because I cannot accurately express the changes I want to make to you in English.

goran-w:
Of course you can. Try me... 😉

love-linger:
> Of course you can. Try me... 

If I were to state directly what I believe the final version of this PR should look like, it would be as follows:

- Minimal and clear modifications
- Need a new control to handle the display for file mode changes and cleanup codes about it in other classes. It means that the `Commands/Models/ViewModels` will only hold the original data and let the new control to handle the display (include translated tooltip) of file mode change

goran-w:
> * Need a new control to handle the display for file mode changes and cleanup codes about it in other classes. It means that the `Commands/Models/ViewModels` will only hold the original data and let the new control to handle the display (include translated tooltip) of file mode change

Good point!

> I've pushed my implementation of this feature.

Thanks! The commit you refer to is 81844e8361db69ecb3f6050e044624cef0aed7e5.

Your changes in that commit are all good, but not complete:

At least let me have a Preference setting, to instead show the filemode names up front (and show the numbers only within the tooltip). That way BOTH CROWDS can be happy, and the (arguably) more user-friendly alternative will be more readily available! 😉 I can implement this, just let me know which option should be the default. (IMHO, though, the more beginner-friendly option should be the default, since the power-users will know where to look for changing it!)

Worse, your commit does not at all address the parsing issues in Diff. It completely ignores the fact that TypeChanged diffs are not parsed correctly. It also does not take my improvements into account at all - I found the original parsing code a bit hard to read and understand, so I split it into 3 main parts (diff-header, chunk-start and chunk-body) and fixed the issues I saw. Also using named constants makes the code easier to read and maintain! (You seem to have not bothered to even understand my changes, or at least ask me to motivate them and discuss them further.) I have now rebased and force-pushed my PR branch, to contain only the remaining refactor + fix. Maybe let me split these changes into a few separate commits, to make them easier to digest, and then we can discuss them in more detail?

Here's how the current Diff-parsing fails (miserably) on a type-changed file (see my original screenshot above, from the same commit, to see how it should look when being more correctly parsed) :

image

love-linger:
That's why I want people to create minimal and clear modifications.

It's very hard to read what's the problem you want to solve from your code. No `git` command outputs.

goran-w:
> That's why I want people to create minimal and clear modifications.
> It's very hard to read what's the problem you want to solve from your code.

Fair enough. Let me try to break my changes up into smaller ones. (Part of it was a need to change the whole flow of parsing, though, hence the larger changes.) So, let me 'regroup' and get back to you here when I'm finished - it will take some time, since I have my 'day job' to take care of. 😅

> No `git` command outputs.

I suppose you mean you want to see examples of the Git diff-command outputs for these cases? I can provide that too...

love-linger:
> At least let me have a Preference setting, to instead show the filemode names up front (and show the numbers only within the tooltip). That way BOTH CROWDS can be happy, and the (arguably) more user-friendly alternative will be more readily available! 

As I said before, I do not like this way. Tooltip should be used to explain or help user understand the original control.

goran-w:
> Tooltip should be used to explain or help user understand the original control.

Ideally, the original control should be easy to understand without a tooltip, and the tooltip should be used to provide addtitional details. Currently the original control displays an arguably less user-friendly (long) 'magic number' instead of a clear-text name.

At least some of your user-base will think that the name is better, and that we shouldn't have to resort to the tooltip to see the clearer representation - instead the tooltip should provide the less prominent info. (I consider myself a 'power-user' by now, but I STILL prefer the names before the numbers... Also, it's not like the people who are used to the numbers will FAIL to understand the names?)

Instead of catering ONLY to one side of the user-base, please let us at least have the choice!

If you REALLY don't want the Preference setting, then PLEASE recognize that EVERYONE will understand the names, but only SOME will understand the numbers! Therefore, prioritizing the names before the numbers should be the 'best' choice. (The numbers will be there in the tooltip, for anyone who wants to check the details.)

goran-w:
@love-linger Thank you for merging my refactoring and fix of the diff-parsing! Much appreciated. 💯👍

Let me know whether I could make a new PR for adding the suggested Preference-setting, or you are able to reconsider putting the names (which everyone can understand) front-and-foremost, instead of the numbers (which only some people readily understand and have memorized).

goran-w:
@love-linger After your code review commit (nice changes, btw!) you might want to make an 'early exit' by adding a `return` after detecting `SPECIAL_BINARY` on lines 148-150 in `Commands/Diff.cs`:

```
if (line.StartsWith(SPECIAL_BINARY, StringComparison.Ordinal))
{
    _result.IsBinary = true;
    return;
}
```

Alternatively, the relevant lines could be moved back to their previous location at the bottom of the function, where no `return` is needed. The reason I had put these lines at the bottom was that they actually appear in that order within the Git diff output (i.e. after the parsing of the `index` line). Example:

```
diff --git ...
index ...
Binary ...
```

goran-w:
> Alternatively, the relevant lines could be moved back to their previous location at the bottom of the function

For reference, this was fixed in commit 7de831f6af636a06d020b1ccabdcf6abff5978f7.

PR #2478, opened 2026-06-24
fix: focus not following default selection in deal-with-local-changes dialog

JC-Chung:
When "Stash & reapply changes by default" is enabled in preferences,
the second radio button is pre-selected but keyboard focus remains on
the first option, causing a mismatch between the focus indicator and
the selected item.

Fix by overriding `OnLoaded` in `DealWithLocalChangesMethod` to focus
the radio button that matches the current `Method` value after the
control is attached to the visual tree, consistent with how
`Reset.axaml.cs` handles initial focus via `OnLoaded`.

### DIFF
--- a/src/Views/DealWithLocalChangesMethod.axaml.cs
+++ b/src/Views/DealWithLocalChangesMethod.axaml.cs
@@ -1,5 +1,6 @@
 ﻿using Avalonia;
 using Avalonia.Controls;
+using Avalonia.Input;
 using Avalonia.Interactivity;
 
 namespace SourceGit.Views
@@ -21,6 +22,24 @@ namespace SourceGit.Views
         public DealWithLocalChangesMethod()
         {
             InitializeComponent();
+        }
+
+        protected override void OnLoaded(RoutedEventArgs e)
+        {
+            base.OnLoaded(e);
+
+            switch (Method)
+            {
+                case Models.DealWithLocalChanges.StashAndReapply:
+                    RadioStashAndReapply.Focus(NavigationMethod.Tab);
+                    break;
+                case Models.DealWithLocalChanges.Discard:
+                    RadioDiscard.Focus(NavigationMethod.Tab);
+                    break;
+                default:
+                    RadioDoNothing.Focus(NavigationMethod.Tab);
+                    break;
+            }
         }
 
         protected override void OnPropertyChanged(AvaloniaPropertyChangedEventArgs change)


love-linger:
In all popups, the first input control will be focused by default. And user can use tab to navigation between them

love-linger:
See src/Views/PopupDataTemplates.cs

JC-Chung:
Thank you for the explanation. I understand that `PopupDataTemplates` focuses the first input control by default.

In most popups this works well — Pull and Push focus a dropdown, `CreateBranch`/`AddWorktree`/`CreateTag` focus a text input, and Fetch focuses a checkbox (where order doesn't matter). However, in `DealWithLocalChangesMethod`, when "Stash & reapply changes by default" is enabled, the second radio button is pre-selected but the focus indicator remains on the first one — causing a mismatch between the visual selection and keyboard focus.

This is more than a visual issue: since `RadioButton` triggers selection via the Space key, pressing Space with focus on the first option would accidentally change the selection from `StashAndReapply` back to `DoNothing`, which is the opposite of the user's preference.

Overriding `OnLoaded` in `DealWithLocalChangesMethod` to call `Focus(NavigationMethod.Tab)` on the matching radio button correctly realigns focus with the selection. This was verified empirically: after the fix, Tab moves from the second option to the third (instead of first to second), which is the expected behavior when the second option is pre-selected.

love-linger:
When you want to press space key on a RadioButton, I think it means you want to change it. Otherwise, you can press tab to go to another input element...

For example: when opening fetch popup, the `Force override local refs` checkbox is focused as default. As you said, pressing `Space` will uncheck it. There's nothing wrong in my view.

JC-Chung:
因為 checkbox 之間不會互相影響，就如同 dropdown 和 text input 不影響 radio button ，但是 radio button 卻是會改變結果。

主要的點還是在於偏好設定選擇了第二個選項作為預設（實際上也有改變），但同時畫面上三個選項各自顯示不同樣式（focus、選取、未選取），而不是選取與關注（focus）在同一個選項上。

又如您提到的使用 Tab 鍵切換選項，常用前兩個選項的話應該會調整偏好設定，所以我想會用到切換最可能是選項三 Discard，但以目前狀況會總要使用兩次 Tab (1 -> 3)（因為不是限定在三個選項中所以也不能用Shift+Tab切換）

如果您覺得維持不更動的話，可以將此PR關閉，因為我已經把我想傳達的都提出來了。

再麻煩您考慮了，謝謝。

love-linger:
> 因為 checkbox 之間不會互相影響，就如同 dropdown 和 text input 不影響 radio button ，但是 radio button 卻是會改變結果。

但是 checkbox 自己也会改变结果呀。

JC-Chung:
The difference is that with a checkbox, the focused state and the effect of pressing Space are visually consistent — the focused checkbox is clearly the target of the action. With a RadioButton group, when focus is on option 1 but option 2 is selected (due to user preference), pressing Space silently overrides the user's preference setting, which contradicts what the UI visually shows as the current selection.

Also, `Reset.axaml.cs` already uses the same `OnLoaded` pattern to manage focus, so this approach is consistent with existing code. The change is limited to a single file and doesn't affect any other popup.

love-linger:
`Reset.axmaml.cs` is just one place where I forget to cleanup...

JC-Chung:
> 但是 checkbox 自己也会改变结果呀。

改變的是當前 `focus` 目標的狀態，但 `RadioButton` 不是。
以對各個類型的控制項點擊 `Space` 來說
| Control 類型 | Space 行為 |
|---|---|
| `TextBox` | 插入空白（`BranchName`/`TagName` 插入 `-`） |
| `ComboBox` | 開啟選單，不會改變值 |
| `CheckBox` | 切換當前 `focus` 項目的值 |
| `RadioButton` | 改變選取，與視覺相反 |

`RadioButton` 的情況下，使用者已透過偏好設定選擇了第二個選項，但 `focus` 在第一個，按 `Space` 會把選擇改回第一個，直接反轉偏好設定的結果。

love-linger:
我提交了我对于该功能的实现，请测试最新的Develop分支

PR #2479, opened 2026-06-24
ux: use tabular numbers for commit time

ChiahongHong:
This PR enables tabular numbers for the `COMMIT TIME` column in the histories view to improve readability, so timestamp digits use consistent widths and align more cleanly across rows.

I believe this helps users who do not want to use a monospace font for the entire interface, but still want numbers to align nicely. Unfortunately, tabular numbers do not apply well to mixed alphanumeric content, so `SHA` cannot benefit from the same approach.

I don't think anyone will miss the "misaligned look" XD

## Before

The red line is a vertical alignment guide. You can see that `17:11:07` is noticeably narrower because it contains `11`.

## After

P.S. `HEAD` is bold, so it does not align with the other rows even with tabular numbers enabled.

### DIFF
--- a/src/Views/Histories.axaml
+++ b/src/Views/Histories.axaml
@@ -218,6 +218,7 @@
               <DataTemplate DataType="m:Commit">
                 <Border Padding="8,0" Background="Transparent" ClipToBounds="True">
                   <v:CommitTimeTextBlock FontWeight="{Binding IsCurrentHead, Converter={x:Static c:BoolConverters.IsBoldToFontWeight}}"
+                                         FontFeatures="+tnum"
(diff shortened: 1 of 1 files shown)


love-linger:
I really like this idea. But I don’t know much about font features. What will happen when the font doesn’t support this feature?

ChiahongHong:
Good question. As far as I know, the vast majority of modern fonts support tabular numbers, so I specifically tested this with Red Hat Text from version 3.0 to see what would happen when the feature is not supported: https://github.com/RedHatOfficial/RedHatFont/archive/refs/tags/3.0.0.zip

As shown in the screenshot, it simply has no effect. It does not seem to cause any side effects, so I believe Avalonia UI handles the unsupported font feature correctly.

PR #2516, opened 2026-07-05
Add day-of-week and localized-date display options

teras:
## Why

Sometimes you need to see the **weekday** of a commit, not just the date — especially when reasoning about workdays, where knowing it was e.g. a Friday matters as much as the calendar date itself. The current display only shows the date.

Also, while the app interface can be localized, the date is always rendered in English, which doesn't help users running the app in their own language. It is preferable to let the date follow the user's language too, when they want it.

Both options are opt-in and were kept deliberately non-intrusive — placed where there is already space and where they make sense in Preferences.

## What

**Day of week** — a dropdown in Preferences ▸ General (None / Short / Full) that prepends `ddd`/`dddd` to the active date format. Stored as `DayOfWeekStyle`; day names come from the active culture.

**Localized dates** — a toggle (`UseLocalizedCulture`) that ties day/month names to the application's selected language instead of the invariant/English culture. The culture is derived from the app locale key, not the OS culture.

Both flow through `DateTimeFormat.Format` and are bound into `CommitTimeTextBlock` and `DateTimePresenter`. New locale keys added to `en_US` and `el_GR`.

### DIFF
--- a/src/App.axaml.cs
+++ b/src/App.axaml.cs
@@ -239,6 +239,7 @@ namespace SourceGit
 
             var pref = ViewModels.Preferences.Instance;
             SetLocale(pref.Locale);
+            Models.DateTimeFormat.UseCulture(pref.Locale);
             SetTheme(pref.Theme, pref.ThemeOverrides);
             SetFonts(pref.DefaultFontFamily, pref.MonospaceFontFamily);
         }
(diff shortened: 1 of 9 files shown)


HermesSantos:
The weekdays are initially set correctly. However, when I change the app language, the weekday names are not updated to the new language. They only refresh after I reselect the weekdays or restart the app.

teras:
Indeed that's true, this part was missing. Now it's fixed.

HermesSantos:
It works as it should now.

PR #2520, opened 2026-07-07
Add .tss to .gitignore

ghiboz:
With visual studio 2026 two .tss files are generated: should be removed from the git flow


love-linger:
I'm using VS2026 and I have not seen such files.

ghiboz:
image
image
I always got these 2 files _(also in other projects with avalonia)_

Issue #985, opened 2025-02-16
Allow fulltext search across all fields (SHA hash, title, author name...)


mdindoffer:
# What
Currently Sourcegit can search for commits in various fields: commit SHA hash, commit message, author, etc. But each of these search fields (or "categories") must be selected individually before initiating the search.

![Image](https://github.com/user-attachments/assets/f1d9b83d-c0c0-4060-9397-cb01435bc5f0)

I would like Sourcegit to have an option to perform a "fulltext" search that performs the search on all of the fields at the same time.

# Why
I often switch the fields I search by. Sometimes I want to search by "Author" to look for commits made by a colleague, other times I want to search for a specific term in all commit messages. Each time I want to search for something else I have to select the corresponding field in the dropdown, which is unnecessary and a little bit annoying. If the search would automatically search all of the fields I would not have to change the settings back and forth all the time.

love-linger:
The GIT commands used behind these 5 ways are completely different, and I haven't found any that support the `OR` operation.

- SHA: `git show --no-show-signature --decorate=full --pretty=format:%H%n%P%n%D%n%aN±%aE%n%at%n%cN±%cE%n%ct%n%s -s <commit_hash>`
- Author: `git log -1000 --date-order --no-show-signature --decorate=full --pretty=format:%H%n%P%n%D%n%aN±%aE%n%at%n%cN±%cE%n%ct%n%s -i --author=<author>`
- Committer: `git log -1000 --date-order --no-show-signature --decorate=full --pretty=format:%H%n%P%n%D%n%aN±%aE%n%at%n%cN±%cE%n%ct%n%s -i --committer=<committer>`
- File: `git log -1000 --date-order --no-show-signature --decorate=full --pretty=format:%H%n%P%n%D%n%aN±%aE%n%at%n%cN±%cE%n%ct%n%s -- <pathspec>`
- Messsage: a litte bit complex. First input string will be split into words, and joined with ` --grep="<escaped_word>"` format, and then use `git log -1000 --date-order --no-show-signature --decorate=full --pretty=format:%H%n%P%n%D%n%aN±%aE%n%at%n%cN±%cE%n%ct%n%s --all-match -i <joined_string>`

mdindoffer:
Well, if git does not support this out of the box, you can always just perform individual searches and then merge the results.

love-linger:
> Well, if git does not support this out of the box, you can always just perform individual searches and then merge the results.

It will take too much time

PR #2543, opened 2026-07-17
feature: add fulltext commit search across SHA/author/committer/message (implements #985)

mdindoffer:
# Summary
This PR adds a "fulltext" search capability to perform a unified search across multiple commit fields at once, which was refused a year ago in #985 due to time constraints.

# How & why
Git has no bult-in logical "OR" `--grep`/`--author`/`--committer` in a single invocation, so the new `All` method (now the default) runs the field queries separately, plus an exact-SHA lookup, and merges the deduplicated results sorted by commit time. Adds an internal-only `ByCommitter` query mode to power it.

# Disclaimer
This was done with the help of Claude, as I have not done any C# development in the last 8 years. Feel free to reject on maintainability concerns.

### DIFF
--- a/src/Commands/QueryCommits.cs
+++ b/src/Commands/QueryCommits.cs
@@ -28,6 +28,10 @@ namespace SourceGit.Commands
             {
                 builder.Append("-i --author=").Append(filter.Quoted());
             }
+            else if (method == Models.CommitSearchMethod.ByCommitter)
+            {
+                builder.Append("-i --committer=").Append(filter.Quoted());
+            }
             else if (method == Models.CommitSearchMethod.ByMessage)
             {
                 var words = filter.Split([' ', '\t', '\r'], StringSplitOptions.RemoveEmptyEntries);
--- a/src/Models/Commit.cs
+++ b/src/Models/Commit.cs
@@ -4,13 +4,18 @@ using CommunityToolkit.Mvvm.ComponentModel;
 
 namespace SourceGit.Models
 {
+    // Bound by index to the search-method ComboBox in Views/Repository.axaml, so
+    // UI-selectable methods must stay first and in dropdown order. ByCommitter is
+    // internal-only (used by the fulltext `All` search).
     public enum CommitSearchMethod
     {
-        BySHA = 0,
+        All = 0,
+        BySHA,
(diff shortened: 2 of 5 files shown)


mdindoffer:
@love-linger Are you accepting pull requests? I tested this manually on my local build, I did not find any issues. What do you think about calling this option "All" if it does not include the path/content search? Should I rename it to something like "Metadata" instead?

love-linger:
Sorry for the delayed reply. I do not think we need this feature. 

- SHA/Author/Message/Path are totally deferent things. Usually, we are not unclear about the types that need to be searched.
- Doing this is unhelpful in terms of both time‑consumption and contribution to the results.

mdindoffer:
Could you please elaborate on why do you think this is "unhelpful"?

I would argue this is a very nice usability improvement, for the justification see the original issue #985 .
You don't have to select what you want to search for, you just type your prompt. This is removing a huge annoyance. That's how search works in almost any software nowadays - Google indexes everything, Gitlab has also unified search for SHAs and content. In my internet banking website, when I'm looking at my account transactions, I don't have pick whether I want to search by IBAN/account number, or by the Name of the partner - I just type it into the search box and the IB app is smart enough to figure it out. Picking your search mode first is an artefact from the 1990s.

When you are searching for multiple "things" in succession you would have to always pick and select the appropriate item from the dropdown, now you finally have a fulltext search where you can copy&paste the commit SHA and the author email and so on.

The performance is OK, I tested this on a repository with thousands of commits, it was instantaneous.

Saving 2-3 mouse clicks per user interaction is a usability design win. When you look for UX best practices for implementing search functionality, they always mention a unified index is a cornerstone:
* https://whatifdesign.co/feeds/blog/best-search-ux-design
* https://moodymedia.io/blog/search-ux-best-practices/
* https://www.coveo.com/blog/search-box-ux-examples/

love-linger:
I think I have explained it very clearly - it will take too long time. For example: with your PR, seaching a simple string `leo` in Unreal Engine repo will takes 10+ seconds (on AMD Ryzen 9 7950X + SSD). While only searching `leo` with `AUTHOR` or `MESSAGE` will only take less than 5 seconds.

PR #2556, opened 2026-07-22
fix: prevent leftmost x-axis label from being clipped in Statistics chart

JC-Chung:
The leftmost x-axis label in the Statistics chart can be partially hidden
when its left edge extends beyond the Y-axis boundary.

This PR fixes the issue by drawing labels outside the clip region,
so they are no longer subject to the `leftMargin` boundary.

image
image


JC-Chung:
Resolved conflicts and updated to match the latest changes.

PR #2568, opened 2026-07-26
fix: prefer exact match over suffix match when auto-selecting tracking branch for worktree

weiningwei:
When adding a worktree with an existing local branch and enabling "Set Upstream Tracking Branch", the `AutoSelectTrackingBranch()` method used `EndsWith` to match remote branches against the local branch name. This caused unintended matches — for example, local branch `smart_build` would match `fix_smart_build` instead of the exact match `smart_build`.

This commit fixes the logic by:
1. First attempting an **exact match** (`Equals`) against remote branch names
2. Falling back to the original **suffix match** (`EndsWith`) only if no exact match is found
3. Keeping `RemoteBranches[0]` as the final fallback


weiningwei:
Sorry, I selected the wrong target branch.

PR #2569, opened 2026-07-26
fix: prefer exact match over suffix match when auto-selecting tracking branch for worktree

weiningwei:
When adding a worktree with an existing local branch and enabling "Set Upstream Tracking Branch", the `AutoSelectTrackingBranch()` method used `EndsWith` to match remote branches against the local branch name. This caused unintended matches — for example, local branch `smart_build` would match `fix_smart_build` instead of the exact match `smart_build`.

This commit fixes the logic by:
1. First attempting an **exact match** (`Equals`) against remote branch names
2. Falling back to the original **suffix match** (`EndsWith`) only if no exact match is found
3. Keeping `RemoteBranches[0]` as the final fallback

### DIFF
--- a/src/ViewModels/AddWorktree.cs
+++ b/src/ViewModels/AddWorktree.cs
@@ -137,7 +137,8 @@ namespace SourceGit.ViewModels
             var name = GetBranchName(true);
             if (!string.IsNullOrEmpty(name))
             {
-                var remoteBranch = RemoteBranches.Find(b => b.Name.Equals(name, StringComparison.Ordinal) || b.Name.EndsWith("/" + name, StringComparison.Ordinal));
+                var remoteBranch = RemoteBranches.Find(b => b.Name.Equals(name, StringComparison.Ordinal));
+                remoteBranch ??= RemoteBranches.Find(b => b.Name.EndsWith("/" + name, StringComparison.Ordinal));
                 if (remoteBranch != null)
                 {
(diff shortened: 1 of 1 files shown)


love-linger:
`EndsWith` is used instead of `Equals` for the following reason: branch names may contain subpaths marked with `/`, such as `feature/xxxx`. When creating a worktree, users do not necessarily need to create the directory named `feature` and may directly use paths like `../xxxx`.

I suggest refactor it to:

```
b.Name equals name or b.Name endWith `/{name}`
```

weiningwei:
Thank you for the explanation. That makes sense — EndsWith("/" + name) prevents the fix ↔ hotfix issue while still handling path-component names like feature/fix. I've updated the PR to use your suggested condition:
`b.Name.Equals(name, StringComparison.Ordinal) || b.Name.EndsWith("/" + name, StringComparison.Ordinal)`.
This replaces the old EndsWith(name) call in AutoSelectTrackingBranch().

Also, would you prefer the two conditions to be split into separate Find calls so that an exact match always takes priority over a suffix match regardless of list order? Currently as a single Find the first match in the list wins.

love-linger:
Good idea. ‘Equal’ first and then ‘EndsWith’

weiningwei:
Updated as suggested — split into two Finds so exact match takes priority:

PR #2600, opened 2026-08-01
localization: complete German translations

mowi12:
Completed de_DE from 89.52% to 100% (107 keys).

### DIFF
--- a/src/Resources/Locales/de_DE.axaml
+++ b/src/Resources/Locales/de_DE.axaml
@@ -25,10 +25,16 @@
   <x:String x:Key="Text.AIAssistant.Model" xml:space="preserve">Modell</x:String>
   <x:String x:Key="Text.AIAssistant.Regen" xml:space="preserve">NEU GENERIEREN</x:String>
   <x:String x:Key="Text.AIAssistant.Tip" xml:space="preserve">Verwende AI, um Commit-Nachrichten zu generieren</x:String>
+  <x:String x:Key="Text.AIAssistant.Use" xml:space="preserve">Übernehmen</x:String>
   <x:String x:Key="Text.App.Hide" xml:space="preserve">SourceGit minimieren</x:String>
+  <x:String x:Key="Text.App.HideOthers" xml:space="preserve">Andere ausblenden</x:String>
   <x:String x:Key="Text.App.ShowAll" xml:space="preserve">Alles anzeigen</x:String>
+  <x:String x:Key="Text.Apply.3Way" xml:space="preserve">3-Way Merge</x:String>
   <x:String x:Key="Text.Apply.File.Placeholder" xml:space="preserve">Wähle die anzuwendende .patch-Datei</x:String>
   <x:String x:Key="Text.Apply.IgnoreWS" xml:space="preserve">Ignoriere Leerzeichen-Änderungen</x:String>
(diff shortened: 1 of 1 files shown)


src/Resources/Locales/de_DE.axaml
```diff
@@ -124,21 +139,29 @@
...
   <x:String x:Key="Text.CommitCM.Checkout" xml:space="preserve">Commit auschecken</x:String>
   <x:String x:Key="Text.CommitCM.CherryPick" xml:space="preserve">Diesen Commit cherry-picken</x:String>
   <x:String x:Key="Text.CommitCM.CherryPickMultiple" xml:space="preserve">Mehrere cherry-picken…</x:String>
   <x:String x:Key="Text.CommitCM.CompareWithHead" xml:space="preserve">Vergleiche mit HEAD</x:String>
   <x:String x:Key="Text.CommitCM.CompareWithWorktree" xml:space="preserve">Vergleiche mit Worktree</x:String>
   <x:String x:Key="Text.CommitCM.CopyAuthor" xml:space="preserve">Autor</x:String>
+  <x:String x:Key="Text.CommitCM.CopyAuthorTime" xml:space="preserve">Autor-Zeitpunkt</x:String>
   <x:String x:Key="Text.CommitCM.CopyCommitMessage" xml:space="preserve">Commit-Nachricht</x:String>
   <x:String x:Key="Text.CommitCM.CopyCommitter" xml:space="preserve">Committer</x:String>
+  <x:String x:Key="Text.CommitCM.CopyCommitterTime" xml:space="preserve">Committer-Zeitpunkt</x:String>
```
Nietod, line 164: Wrong. That should be "Commit-Zeitpunkt" because we are in the context menu of the history where the table head says "Commit Time" -> "Commit-Zeitpunkt" and we are talking about the same timestamp. Refering to the Committer makes no sence there.

Therefore the English base translation is wrong.

mowi12: I'm not sure about this.

I would agree with you that it makes no sense to reference the committer time, since there's only the time when a commit was created.

What solution would you propose?
Should we change all translations, including the variable name, to `Text.CommitCM.CopyCommitTime`?
Or should we simply keep `Committer-Zeitpunkt` as the translation, since all the other languages also refer to it that way?

gadfly3173: This translation is based on Git's own definition and can be understood as setting the author date (Author Date) and the committer date (Committer Date). Reference: https://git-scm.com/docs/git-commit#_commit_information

mowi12: So, if I understand correctly, @gadfly3173, you would keep the current translation?

gadfly3173: My view is that “Committer Date/Time” is the correct translation, and this is consistent with the Git documentation. However, I am not a maintainer, so the final decision rests with the maintainer @love-linger .

mowi12: Great, thanks!

Nietod: @gadfly3173 got a good point there! `Committer-Time` is technically correct and `Commit-Time` would be more UX friendly.

Happy with either solution and @love-linger can decide, agreed!

love-linger:
I believe both translations will not cause confusion for users. Either is acceptable.

PR #2604, opened 2026-08-03
feat: workspace lookup before opening from CLI

gitPhate:
This PR improves the command line repo opening by looking up the repository among the ones already opened in all workspaces.
Tests done on win 11, with 3 workspaces configured:
1. Program closed → CLI-open repo in a different workspace — works
2. Program closed → CLI-open a different repo already in the same workspace — works
3. Program open → CLI-open repo in a different workspace — works
4. Program open → CLI-open a different repo in the same workspace — works
5. Multi-tab target workspace — CLI-open correctly focused the right tab within the target workspace, not just whatever tab was last active there
6. Bare repository, using the test repo at C:\Development\test-bare-repo.git — switched workspace correctly, tested both closed and already-open app states
7. Case-insensitive path (CLI invoked with a lowercase path against a repo added with different casing) — matched correctly
8. Persistence after CLI switch — relaunching normally afterward opens into the workspace the CLI switch left you in, as expected

### DIFF
--- a/src/ViewModels/Launcher.cs
+++ b/src/ViewModels/Launcher.cs
@@ -87,6 +87,7 @@ namespace SourceGit.ViewModels
                 {
                     var node = Preferences.Instance.FindOrAddNodeByRepositoryPath(repo, null, false);
                     Welcome.Instance.Refresh();
+                    TrySwitchToContainingWorkspace(node);
                     OpenRepositoryInTab(node, null);
                     return true;
                 }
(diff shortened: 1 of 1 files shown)


love-linger:
This PR makes it impossible to share a repository in multi-workspaces.

gitPhate:
If one repo is opened in more than one workspace it would open the first match. Can you tell me what it would be the desired behaviour in this case?
We can maybe use a CLI flag to trigger this behaviour, leaving the default to how it was before.

gitPhate:
@love-linger do you have an update on this one?
I'll tell you why I'mma need this change: I'd like to bind a key combination to the opening of a specific repo in sourcegit, so I can quickly access it from the keyboard instead of navigating in tabs. Since I professionally work as a developer I have a ton of repositories opened inside sourcegit, it would be quite useful to me to have this option.

PR #2612, opened 2026-08-07
feat: Font Selector

ennerperez:
image

- Font Selector based on OS fonts

### DIFF
--- a/src/ViewModels/Preferences.cs
+++ b/src/ViewModels/Preferences.cs
@@ -1,10 +1,12 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 using System.Threading.Tasks;
 using Avalonia.Collections;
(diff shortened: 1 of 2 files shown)


love-linger:
Not planned. In fact, early versions of SourceGit (such as v8.8) were implemented this way. However, due to certain issues with font‑handling in AvaloniaUI, SourceGit had to support fallback‑style settings such as `Consolas, Microsoft YaHei UI`. 

image

ennerperez:
Yes, I do remember. For that reason, this implementation is simpler, just the font names without rendering the font in the dropdown.

love-linger:
It's not just the performance issue of font selector. We must support settings like `Consolas, Microsoft YaHei UI`

appleneko2001:
> It's not just the performance issue of font selector. We must support settings like `Consolas, Microsoft YaHei UI`

its not hard to implement something like List and add a basic sort button for multiple font family option
The performance was affected by font preview rendering mostly

PR #2613, opened 2026-08-07
feat(launcher): open submodule tabs next to parent with prefixed titles

berkanuslu:
Submodule tabs now insert right after the parent repo tab instead of appending at end, and display a `Parent : Submodule` title for quick identification.

## Changes

- **Tab placement** — Pass parent page hint through `OpenRepositoryInTab` so submodule tabs insert at `parentIdx + 1`
- **Title prefix** — Compose `"ParentName : SubfolderName"` for submodule tab titles using a single `ComposeSubmoduleTitle` helper
- **Live rename sync** — Recompose title on every open so renaming the parent immediately reflects in child tabs

## Before / After

| Before | After |
|--------|-------|
| A, B, C, D → open submodule in B → A, B, C, D, **BB** | A, B, C, D → open submodule in B → A, B, **B : BB**, C, D |

## Files Changed

- `src/ViewModels/Launcher.cs` — tab insertion logic + title composition
- `src/ViewModels/Repository.cs` — pass parent page to launcher

### DIFF
--- a/src/ViewModels/Launcher.cs
+++ b/src/ViewModels/Launcher.cs
@@ -296,7 +296,7 @@ namespace SourceGit.ViewModels
                 IsUnmanaged = true
             };
 
-            OpenRepositoryInTab(node, null);
+            OpenRepositoryInTab(node, page);
         }
 
         public void OpenRepositoryInTab(RepositoryNode node, LauncherPage page)
@@ -353,6 +353,13 @@ namespace SourceGit.ViewModels
                     page.Node = node;
                     page.Data = repo;
                 }
+            }
+            else if (page.Node.IsRepository)
+            {
+                var parent = page;
+                page = new LauncherPage(node, repo);
+                var parentIdx = Pages.IndexOf(parent);
+                Pages.Insert(parentIdx + 1, page);
             }
             else
             {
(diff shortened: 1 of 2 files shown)


src/ViewModels/Launcher.cs
```diff
@@ -468,6 +479,12 @@ private void PostActivePageChanged()
             CommandPalette = null;
         }
 
+        private static string ComposeSubmoduleTitle(LauncherPage parent, RepositoryNode node)
+        {
+            var baseName = Path.GetFileName(node.Id);
+            return $"{parent.Node.Name} : {baseName}";
```
WizardCM:
```suggestion
            return $"{parent.Node.Name}: {baseName}";
```
This looks less strange in English.

berkanuslu: That's fair :)

love-linger:
I've pushed my implemetation for this feature.

- Create another method `OpenSubRepository` instead of modify the existing one `OpenRepositoryInTab` because some calculation is not necessary for submodules.
- It will open submodule next to the owner page
- It will set the name to `<owner_pure_name>: <submodule>` by default. For example: when open a nexted submodule `C` from another submodule `A: B`, the tab's name will be `B: C` instead of `A: B: C` by default. But after you edit the tab name manually, `SourceGit` will use it instead of the default one.

PR #2619, opened 2026-08-11
Improve inline-diff highlighting

goran-w:
This PR does two simple changes to make the highlighting of inline diffs more consistent and useful:
* Show inline changes even for diff-(c)hunks with unmatched delete/add counts.
  * We now try to match the first N lines changed on both sides of the diff, instead of giving up just because the count is different.
* Increase the limit of how many inline-chunks are allowed per line. (Also use named constants.)
  * The current limit was way too restrictive. Two similar (and perfectly ordinary) changed lines could easily fall into one being inline-highlighted and the other not.
  * A lot of work would already have been performed to calculate these inline-chunks, only to throw them away above this very restrictive limit.

### DIFF
--- a/src/Commands/Diff.cs
+++ b/src/Commands/Diff.cs
@@ -35,6 +35,9 @@ namespace SourceGit.Commands
         private const string SPECIAL_NO_NEWLINE = " No newline at end of file";
         private const string SPECIAL_SUBMODULE = "Subproject commit ";
 
+        private const int MAX_INLINE_CONTENT_LENGTH = 1024;
+        private const int MAX_INLINE_CHUNKS_PER_LINE = 16;
+
         public Diff(string repo, Models.DiffOption opt, int numContextLines, bool ignoreWhitespace, bool ignoreCRAtEOL)
         {
             _result.TextDiff = new Models.TextDiff();
@@ -330,11 +333,11 @@ namespace SourceGit.Commands
                         var left = _deleted[i];
                         var right = _added[i];
 
-                        if (left.Content.Length > 1024 || right.Content.Length > 1024)
+                        if (left.Content.Length > MAX_INLINE_CONTENT_LENGTH || right.Content.Length > MAX_INLINE_CONTENT_LENGTH)
                             continue;
 
                         var chunks = Models.TextInlineChange.Compare(left.Content, right.Content);
-                        if (chunks.Count > 4)
+                        if (chunks.Count > MAX_INLINE_CHUNKS_PER_LINE)
                             continue;
 
                         foreach (var chunk in chunks)


love-linger:
I don't think this will be better

image

This is the diff in GitHub:

image

You can see that GitHub also ignore inline-diff when the count of deleted is mismatch the count of added.

goran-w:
@love-linger It's of course a bit hit-or-miss, but there are many VERY TRIVIAL cases where a "mismatched" inline-diff would make a lot of sense, and would otherwise miss the inline-diff entirely.

For example:
* extra empty line(s) added/removed below changed line(s)

image

* extra code added/removed below changed line(s)

image

* code below the first changed line(s) changed extensively but the first line(s) have trivial changes

I can easily see past some "false positives" if I get to see these very useful (trivial) ones!

NOTE: The "matched number of lines" inline-diffs are sometimes just as "false" as the "mismatched" ones, so there's actually not much of a difference in these two cases. (The only way to make both cases better would be to implement a deeper and more "expensive" search to find the most similar lines in the diff-hunk. But for now, this change at least makes sure we don't miss the most trivial cases of mismatched lines.)

(Also, just because GitHub doesn't do this, it doesn't mean their implementation is the better one...)

goran-w:
@love-linger Also, to motivate my other commit in this PR (where I increased the limit on the number of inline-diffs per line), here are two trivial examples where the earlier existing implementation skips showing the (already calculated) inline-diffs:

* on one of two changed lines

image

* on both changed lines

image

But with my proposed change, both of these cases are handled more correctly:

* on one of two changed lines

image

* on both changed lines

image

love-linger:
The original design intention of inline‑diff is to highlight modified parts when the changed content is minor compared with the original content (which may be hard to spot); when the changes are extensive, amounting to practically a full rewrite, no highlighting will be applied (the entire line will be directly treated as modified). The current solution does have certain issues, since code cannot identify differences at a glance the way humans can. For example:

image

You can see the 3 lines have been totally rewrited. We should not use inline-diff highlights here.

goran-w:
> You can see the 3 lines have been totally rewrited. We should not use inline-diff highlights here.

I beg to differ! I'd gladly take these "false" highlights, as long as the trivial highlights from my examples are covered! This is as good we can make it without a more extensive and expensive algorithm.

There's a pretty large risk of not spotting small/minor changes, when the inline-diff highlights SOME of them but not ALL of them, as all my example screenshots above show.

love-linger:
You may have misunderstood this feature: No hightlights does not mean no changes. Instead, it means the change is large!!!

goran-w:
> You may have misunderstood this feature: No hightlights does not mean no changes. Instead, it means the change is large!!!

I completely understand that. 🙄 BUT if you look closely at my examples, the minor changes could easily be non-spotted JUST BECAUSE the overall changes are larger. Also, some of the "large" changes in my examples are NOT large, they are in fact very small but still cause mismatched lines (as in the example of an added empty line).

And the limit on the number of inline-chunks causes VERY SIMILAR changed lines to HAVE or NOT HAVE highlights - which even seems random at first! A coworker asked me about this and was very confused, which led me to implement this PR...

Additional suggestion: add a toggle "Highlight inline diffs" so we can quickly turn the feature on/off. (Similar to the existing "Show hidden symbols" toggle.) That way, we can allow the inline-diff to cover these cases, while also allowing it to be easily toggled off if/when it "gets in the way".

love-linger:
Why do you think the examples you gave are common scenarios? Even if we increase the number of change blocks displayed per line, is it necessary to jump directly from 4 to 32? Can changes with more than 10 blocks in a single line really be regarded as minor changes?

love-linger:
image

In fact, the example above is precisely a special case. For us humans, it is certainly easy to understand that what remains unchanged is the overall structure of the XML node, while what has changed are the values of the attributes. However, it can also be interpreted that the new position/rotation is completely different from the old one, since they have lots of differences (not minor modifications).

goran-w:
> Why do you think the examples you gave are common scenarios? 

They are all real-world scenarios, from our codebase (.xml files) and from SourceGit (.md file). The only "contrived" example is the one where I inserted a new XML attribute "a" and changed the indentation, to quickly create a case where both lines would lack highlights due to the number of inline-chunks.

> Even if we increase the number of change blocks displayed per line, is it necessary to jump directly from 4 to 32? Can changes with more than 10 blocks in a single line really be regarded as minor changes?

Yes, the increase needs to be substantial. These individual blocks can be trivial, they sometimes include whitespace and indentation etc. XML / XAML examples are telling, there could be many small changes to separate attributes, and highlighting the changed values help a lot in reading the diff.

It's rather frustrating if two similar changed lines make a different decision of highlighted vs not. (And the work of calculating these changes is already done, so there's very little performance impact in actually showing them.)

> it can also be interpreted that the new position/rotation is completely different from the old one, since they have lots of difference

So what? Of course I understand that the whole pos/rot is changed, but the highlights still help me with readability (similar to syntax coloring, but for spotting the actual changed values). Also, it still makes it much quicker to determine if (for example) 1 out of 4 values is UNCHANGED - but for that to be reliable, we must also be able to trust that the highlighting is not "arbitrarily" disabled/skipped on some lines...

What about my suggestion of adding a toggle for "Highlight inline diffs" (or similar wording)? If that's something to consider, I could look into it (as a separate PR)...

love-linger:
I do think this is the most common scenarios of git users in real-world:

image

You can see that with your PR it highlights blocks anywhere. I know that there's no performance issue with your PR. I just think that hightlighting does not help me to read code any longer.

goran-w:
> You can see that with your PR it highlights blocks anywhere. I know that there's no performance issue with your PR. I just think that hightlighting does not help me to read code any longer.

OK. I force-pushed my PR to now include only the commit with increased limit on inline-chunks, to remove some of the perceived "randomness".

How about adding a toggle for "Highlight inline diffs" (or similar wording)? If that's something to consider, I could look into it (as a separate PR)?

love-linger:
- Move contants to the block next to `private const string SPECIAL_DIFF_START = "diff ";`
- I can only accept increasing `maxChunksPerLine` to 16

goran-w:
> * Move contants to the block next to `private const string SPECIAL_DIFF_START = "diff ";`
> * I can only accept increasing `maxChunksPerLine` to 16

Done.

PR #2622, opened 2026-08-13
ux: improve checkbox contrast in dark theme

a2580vb:
### Purpose

Improve checkbox contrast in dark theme.

#### 1. `src/Resources/Themes.axaml`

Added checkbox-specific resources to both Light and Dark theme dictionaries:

- **`Color.CheckBox.Border`**: Set to `#898989` (same as Border1) in the Light theme, and `#A8A8A8` (a brighter border for better visibility) in the Dark theme.
- **`Brush.CheckBox.CheckedBackground`**: Remains `Transparent` in the Light theme, while using the system accent color in the Dark theme.
- **`Brush.CheckBox.CheckedBorder`**: Remains `#898989` in the Light theme, while using the system accent color in the Dark theme.
- **`Brush.CheckBox.CheckedFG`**: Uses the system accent color (unchanged) in the Light theme, and white in the Dark theme.

Additionally, `Color.CheckBox.Border` is registered as the `Brush.CheckBox.Border` brush in the shared section, and the three checked-state brushes are defined directly as `SolidColorBrush` instances within each theme dictionary.

#### 2. `src/Resources/Styles.axaml`

- **Unchecked border brush**: Changed from `Brush.Border1` to `Brush.CheckBox.Border`.
- **Added checked state style `^:checked`**: Uses themed background and border brushes when checked.
- **Check mark icon**: `Fill` changed from the template default `Brush.Accent` to `Brush.CheckBox.CheckedFG`.
- **`:focus-visible:checked`** style updated to use the same themed resources.

### Result

- **Dark theme**: Checked state now uses an accent-colored background with a white check mark, and a brighter border, significantly improving contrast.
- **Light theme**: All values remain identical to before; no visual change.

### DIFF
--- a/src/Resources/Styles.axaml
+++ b/src/Resources/Styles.axaml
@@ -1021,7 +1021,7 @@
     <Setter Property="Template">
       <ControlTemplate>
         <Grid Background="Transparent" ColumnDefinitions="Auto,*">
-          <Border x:Name="Border" Grid.Column="0" Width="16" Height="16" VerticalAlignment="Center" BorderBrush="{DynamicResource Brush.Border1}" BorderThickness="1" Background="Transparent" CornerRadius="2">
+          <Border x:Name="Border" Grid.Column="0" Width="16" Height="16" VerticalAlignment="Center" BorderBrush="{DynamicResource Brush.CheckBox.Border}" BorderThickness="1" Background="Transparent" CornerRadius="2">
             <Path x:Name="Icon" Height="12" Width="12" Data="{DynamicResource Icons.Check}" Fill="{DynamicResource Brush.Accent}" IsVisible="False" Margin="0,2,0,0"/>
           </Border>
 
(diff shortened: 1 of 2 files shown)


love-linger:
image

This style is used by `:checked:focus-within`

a2580vb:
You might not be using a Windows machine, So I'd like to share some screenshots for reference. On Windows 11, the appearance before this change looks like this:
image
And after the change:
image
A previous commit adjusted the focus behavior, and I've since pushed an updated version that restores the original focus functionality while keeping the contrast fix intact.

love-linger:
It's because you system accent color is gray

For example, when your system accent color is blue:

image

you will get:

image

a2580vb:
Thanks for the explanation! I now understand that the issue was caused by the system accent color being set to gray on my end, rather than an actual rendering bug. I'll close this PR accordingly.
Really appreciate you taking the time to look into this and point out the root cause. It saved me from going down the wrong path.
感谢大佬帮忙排查！原来是系统强调色被设成了灰色导致的，不是真正的渲染问题，我将关闭这个PR。
非常感谢你花时间帮忙定位原因，少走了很多弯路，再次感谢！

PR #2626, opened 2026-08-15
feat: use the same rules as git for branch and tag name validation

### LINKED ISSUE #2624: Branch names containing & are rejected although allowed by git
Minor issue, r&d/test (i'm guessing anything with a & in the branch name) is rejected although it's a valid branch name.

Image

philfontaine:
Fixes #2624

### DIFF
--- a/src/Models/RefName.cs
+++ b/src/Models/RefName.cs
@@ -3,7 +3,7 @@ using System;
 namespace SourceGit.Models
 {
     /// <summary>
-    ///     Validates branch, tag and remote names using the same rules that git itself enforces,
+    ///     Validates branch and tag names using the same rules that git itself enforces,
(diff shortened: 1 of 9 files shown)


src/Models/RefName.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            return IsValidRefName(name);
+        }
+
+        public static bool IsValidRemoteName(string name)
+        {
+            if (string.IsNullOrEmpty(name))
+                return false;
+
+            return IsValidRefName(name);
```
love-linger: SourceGit does not support remote with '/'

philfontaine: I excluded remotes from this PR

PR #2627, opened 2026-08-16
feat: Adding Arch Linux package generation

daedalus-codes:
As an Omarchy Linux user, I wanted to provide a means to generate Arch Linux package on releases.  I'm hoping this may alleviate AUR package maintainers of some manual steps to re-release the package for Arch based distributions.  Additionally, with recent supply chain attacks focused on AUR, I hope this would alleviate any worry about any malicious activity.


daedalus-codes:
Still a WIP - I've been testing the GitHub Actions build locally using [nektos/act](https://github.com/nektos/act) - to hopefully get faster feedback on the workflow build/package process loop.

What I hoped would be a weekend project turned out to be a little bigger.  I've run into some challenges b/c it uses containers that are run as root.  `makpkg` will *not* run if the user is root.

I can clean up my changes and provide my config changes to the workflow `.yml` files in another PR is anyone else would be interested.

daedalus-codes:
I was able to get the package to build properly locally within using nektos/act! I believe that GH actions are run as a non-root user, so this shouldn't be an issue. However, if it is, I should be able to address it quickly.

daedalus-codes:
Confirmed works 💯 locally w/ ACT and using the action runners on my fork!

daedalus-codes:
I have confirmed the actions changes will run - I didn't think to link to it [here](https://github.com/AEthersoftware-Works/sourcegit/actions/runs/32691644869)

I have downloaded the package and installed it & it's running on my machine. 👇👇👇

image

daedalus-codes:
Got some notifications about it not passing, will look into it.

daedalus-codes:
Verified GitHub actions completed [here](https://github.com/AEthersoftware-Works/sourcegit/actions/runs/35739762438)

PR #2631, opened 2026-08-17
feature: add "Commit & Force Push" action when amending a commit

### LINKED ISSUE #2575: Add "Commit & Force Push" action when amending a commit
### Description

When working in **Local Changes**, SourceGit normally provides both:

* `Commit`
* `Commit & Push`

However, when the **Amend** option is enabled, only the `Commit` action remains available.

Since amending an already pushed commit usually requires rewriting the remote branch, it would be useful to provide an additional action:

* `Commit & Force Push`

For safety, this action should preferably use `git push --force-with-lease` rather than an unconditional `--force`.

### Expected behavior

When **Amend** is enabled:

* keep the existing `Commit` action;
* provide a `Commit & Force Push` action;
* optionally display a confirmation dialog explaining that the remote branch history will be rewritten;
* use `--force-with-lease` to avoid overwriting remote changes made by other users.

### Motivation

The current workflow requires two separate operations:

1. amend the commit;
2. manually perform a force push afterward.

Adding this action would make the amend workflow consistent with the existing `Commit & Push` workflow while retaining appropriate safety checks.

Related discussion: #2379

sdpfigueiredo:
Closes #2575

### Description

Implements the requested `Commit & Force Push` action in the **Local Changes** panel, making the amend workflow consistent with the existing `Commit & Push` workflow.

- The **Commit & Push** button is now a **SplitButton** with a dropdown offering two actions:
  - `Commit & Push` — unchanged behavior
  - `Commit & Force Push`
- Selecting `Commit & Force Push` first shows a confirmation dialog warning that the remote branch history will be rewritten.
- After confirmation, the commit is created and pushed via the existing push machinery with `--force-with-lease`, so concurrent remote changes by other users are never overwritten blindly (per the issue's safety requirement).

### Notes

- The dropdown is visible whenever the repository has remotes, so the action works both with and without the **Amend** checkbox — it matters most when **Amend** is enabled, since rewriting an already-pushed commit requires a force push.
- New localization keys were added to `en_US.axaml`; all other locales fall back to it via `ResourceInclude`.

### Testing

- `dotnet build src/SourceGit.csproj -c Debug` — 0 warnings, 0 errors
- `dotnet publish src/SourceGit.csproj -c Release -r win-x64` — Native AOT publish succeeds
- Smoke-tested the published Windows x64 build

### Screenshot
image

### DIFF
--- a/src/Resources/Locales/en_US.axaml
+++ b/src/Resources/Locales/en_US.axaml
@@ -999,11 +999,13 @@
   <x:String x:Key="Text.WorkingCopy.ClearCommitHistories" xml:space="preserve">Clear History</x:String>
   <x:String x:Key="Text.WorkingCopy.ClearCommitHistories.Confirm" xml:space="preserve">Are you sure you want to clear all commit message history? This action cannot be undone.</x:String>
   <x:String x:Key="Text.WorkingCopy.Commit" xml:space="preserve">COMMIT</x:String>
+  <x:String x:Key="Text.WorkingCopy.CommitAndForcePush" xml:space="preserve">COMMIT &amp; FORCE PUSH</x:String>
   <x:String x:Key="Text.WorkingCopy.CommitAndPush" xml:space="preserve">COMMIT &amp; PUSH</x:String>
   <x:String x:Key="Text.WorkingCopy.CommitMessageHelper" xml:space="preserve">Template/History</x:String>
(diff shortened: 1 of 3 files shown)


love-linger:
`Force Push` is a dangerous operation. `SourceGit` requires user to manually push with `--force` enabled.

PR #2632, opened 2026-08-17
ux: allow interrupting in-progress push operation in dialog

### LINKED ISSUE #2603: Push Changes to Remote dialog should be closable (or short timeout should be used)
Hello. First of all. Thank your for product. 
Second, sometimes in the case of down of corporative vpn, I have big lag to push changes to unaccessible remote server like several minues. I suspect internally used very big timeout. In order to improve user exprience I suppose timeout should be reduced (as fast fix) or ability to interrupt pushes should added (like cancel button or whatever)

UPD: my version is v2026.09 under linux ubuntu 22.04

sdpfigueiredo:
Resolve #2603

## Problem

When the remote server is unreachable (e.g. corporate VPN down), the `Push Changes to Remote` dialog blocks for minutes on the hung `git push` process with no way to interrupt it — the dialog is not closable while the operation is in progress.

## Solution

Add an opt-in cancellation mechanism for popups and wire it up for the Push dialog:

- **`Popup`** base class gains `CanCancel` (default `false`) and `Cancel()`, and `Cleanup()` becomes virtual.
- **`Push`** overrides `CanCancel => true` and `Cancel()` → cancels a `CancellationTokenSource` whose token is passed to the push command via the existing `.WithCancellation(token)` extension. `Command.ExecAsync()` already kills the git process tree on cancellation, and returns `true` when cancelled — so the dialog closes itself once the operation is interrupted.
- **`PopupRunningStatus`** shows a **CANCEL** button next to the running indicator, visible only when the popup opts in (`CanCancel`).
- **`LauncherPage.CancelPopup()`** now forwards to `popup.Cancel()` when a cancellable popup is in progress, so **ESC** and **mask-click** also interrupt the running push.

## Behavior

1. Push is running (hung on unreachable remote)
2. User clicks **CANCEL** (or presses ESC / clicks outside)
3. `git push` process tree is killed
4. Dialog closes immediately

## Notes

- Changes are opt-in: no other dialog is affected (button only appears for popups with `CanCancel == true`).
- `PushTag` / `PushRevision` follow the same pattern and can opt in later if desired.

Tested: `dotnet build` clean (0 warnings / 0 errors). Windows `win-x64` NativeAOT publish successful.

## Screenshot
Captura de ecrã 2026-08-17 215930

### DIFF
--- a/src/ViewModels/LauncherPage.cs
+++ b/src/ViewModels/LauncherPage.cs
@@ -107,8 +107,14 @@ namespace SourceGit.ViewModels
 
         public void CancelPopup()
         {
-            if (_popup == null || _popup.InProgress)
+            if (_popup == null)
                 return;
+
+            if (_popup.InProgress)
+            {
+                _popup.Cancel();
+                return;
+            }
 
             _popup?.Cleanup();
             Popup = null;
--- a/src/ViewModels/Popup.cs
+++ b/src/ViewModels/Popup.cs
@@ -34,7 +34,7 @@ namespace SourceGit.ViewModels
                 ProgressDescription = desc;
         }
 
-        public void Cleanup()
+        public virtual void Cleanup()
         {
             _log?.Unsubscribe(this);
         }
@@ -42,6 +42,18 @@ namespace SourceGit.ViewModels
         public virtual bool CanStartDirectly()
         {
             return true;
+        }
+
+        /// <summary>
+        /// Indicates the in-progress operation of this popup can be interrupted.
+        /// </summary>
+        public virtual bool CanCancel => false;
+
+        /// <summary>
+        /// Interrupts the in-progress operation of this popup.
+        /// </summary>
+        public virtual void Cancel()
+        {
         }
 
         public virtual Task<bool> Sure()
--- a/src/ViewModels/Push.cs
+++ b/src/ViewModels/Push.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace SourceGit.ViewModels
@@ -190,6 +191,9 @@ namespace SourceGit.ViewModels
         {
             using var lockWatcher = _repo.LockWatcher();
 
+            _cts?.Cancel();
+            _cts = new CancellationTokenSource();
+
             var remoteBranchName = _selectedRemoteBranch.Name;
             ProgressDescription = $"Push {_selectedLocalBranch.Name} -> {_selectedRemote.Name}/{remoteBranchName} ...";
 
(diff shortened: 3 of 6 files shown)


love-linger:
Not planned yet. See #1012

love-linger:
Currently, we use `Process.Kill(true)` to force terminating a git sub-process. It's unsafe and only used for read-only git commands (such as `git diff`).

If you are running `SourceGit` on Windows, I can push a branch that can gracefully ternimate `git` process.

love-linger:
I've pushed my implementation (only supported Windows)

Mingun:
Sending <kbd>Ctrl+C</kbd> to git process will not interrupt git operation correctly?

love-linger:
> Sending Ctrl+C to git process will not interrupt git operation correctly?

Yes. I just do this on Windows. For Linux/macOS, you can not gracefully terminate a process tree with signal SIGINT/SIGTERM (Process.Start does not spawn process in a seperate session/group)

Mingun:
>  (Process.Start does not spawn process in a seperate session/group)

Does it matter? I was sure, that the `git` process is responsible to correctly handle <kbd>Ctrl+C</kbd> and if it allow that to do, that should be safe.

love-linger:
As I said - Sending Ctrl+C works fine on Windows. On Linux/macOS, there are some issues in sending signal SIGINT/SIGTERM. See https://github.com/dotnet/runtime/issues/44944

aikawayataro:
> > (Process.Start does not spawn process in a seperate session/group)
> 
> Does it matter? I was sure, that the `git` process is responsible to correctly handle Ctrl+C and if it allow that to do, that should be safe.

The way dotnet handles child processes makes it hard to reliably wait for completion. It is safe to kill `git` processes with `SIGINT` and `SIGTERM`, but because signal is not sent to the children, they will become background "zombie" processes for a while. From my point of view, this behavior is totally fine, main `git` process exited and any filesystem IO is cleanly closed, the live children are harmless network fetch processes.
I come up with somewhat elegant workaround here: https://github.com/sourcegit-scm/sourcegit/pull/1012#issuecomment-2683809176 the rest of discussion will tell you why it was not merged.

love-linger:
@aikawayataro I've pushed a new commit that uses `setsid` for cancellable git commands on Linux. But I only tested it on my Ubuntu 24.04. Can you review and test it for me?

aikawayataro:
> @aikawayataro I've pushed a new commit that uses `setsid` for cancellable git commands on Linux. But I only tested it on my Ubuntu 24.04. Can you review and test it for me?

I've tested your change on my Arch Linux system, it works, the whole git process tree is terminated, but I would avoid `Process.Kill` as a fallback path, because it can force-kill git and corrupt `.git` database.

love-linger:
> I've tested your change on my Arch Linux system, it works, the whole git process tree is terminated, but I would avoid Process.Kill as a fallback path, because it can force-kill git and corrupt .git database.

So, what is your suggestion for handling cases when `kill(-pgid, SIGTERM)` fails to execute?

UPDATED:

After checking the `kill` manual`, I find:

image

I've update the code to:

```csharp
public void TerminateProcess(Process proc)
{
    if (kill(-proc.Id, 15) != 0)
    {
        // If the process already exited, we can just ignore the error.
        if (Marshal.GetLastPInvokeError() == 3 /* ESRCH */)
            return;

        // Actually, this will not be called since the process is
        // spawned by us (EPERM will not happen), and SIGTERM (15)
        // is a valid signal (EINVAL will not happen).
        // See https://www.man7.org/linux/man-pages/man2/kill.2.html
        try
        {
            proc.Kill(true);
        }
        catch
        {
            // Ignore any errors when trying to kill the process
        }
    }
}
```

PR #2641, opened 2026-08-20
fix: keep blame metadata aligned while scrolling

gadfly3173:
Blame metadata could drift away from source lines when the first visible commit group started above the viewport. Large groups could also clip their hash at the top or make adjacent labels overlap, while measuring only visible lines could change the margin width during scrolling.

Align metadata with AvaloniaEdit line centers, protect the first group only when there is enough room, and measure all blame data consistently so scrolling keeps labels stable and readable.

Before:

image
image

### DIFF
--- a/src/Views/Blame.axaml.cs
+++ b/src/Views/Blame.axaml.cs
@@ -27,81 +27,196 @@ namespace SourceGit.Views
                 ClipToBounds = true;
             }
 
+            private sealed class CommitInfoLink
+            {
+                public CommitInfoLink(Models.BlameLineInfo info, FormattedText shaLink, double lineCenter)
+                {
+                    Info = info;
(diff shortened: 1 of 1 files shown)


love-linger:
I just pushed a commit that makes this UI easier to understand

gadfly3173:
image

Currently it would just not show the first chunk's metadata.

gadfly3173:
image

love-linger:
Done. See commit 9fb864b24b5cee6f18f08fe6540618f9993d6967

gadfly3173:
正常部分：
image

画面中最高的chunk接近一页时，左侧重复渲染
image
image

这样又出现了一开始要修复的重复/偏移问题呀

Issue #2336, opened 2026-05-12
[FR] :  git cherry feature  


heartacker:
The biggest advantage of GUI is that we can intuitively view abundant information and provide intuitive feedback and guidance in interaction.
It is hoped to implement a path tracking function that can highlight (bold) the associated paths in real time.

This feature is very common in many graph-based software, such as PCB layout, microelectronics layout design, and so on.

love-linger:
In your video, why the `orange` (left most) path before commit `v2026.10` is only highlighted when hover branch `H`? Does commit `v2026.10` is not belongs to other branches?

I actually do not like this feature. You can check out how the `Only hightlight current branch` feature works. It hightlights all the subpaths that belongs to current branch.

heartacker:
> In your video, why the `orange` (left most) path before commit `v2026.10` is only highlighted when hover branch `H`? Does commit `v2026.10` is not belongs to other branches?

This is a problem with my implementation. My video is just for the convenience of expressing my concept.


这个是bug ，只是我的实现问题。我只是抛砖引玉的表达我的概念。

---

> I actually do not like this feature. You can check out how the `Only hightlight current branch` feature works. It hightlights all the subpaths that belongs to current branch.


`Only hightlight current branch` and my branch terminal are two different things. In numerous and complex branches, this one can highlight the commit you are in by bold, making it convenient for you to find related commits and other scenarios.

高亮当前分支和我的这个分支线跟踪两码事。在复杂且繁多的分支视图，这个可以在视觉上很好的引导我们所要操作的分支和关联的commit，他主要是提高用户的体验，快速利用 gui 辨识相关的commit。

<img width="683" height="299" alt="Image" src="https://github.com/user-attachments/assets/ba325b8c-e141-4265-a10d-821b1efc1187" />
   
比如电路图。密密麻麻的时候。鼠标悬浮高亮可以快速让你在鼠标游走的时候。快速判断走向。
<img width="397" height="298" alt="Image" src="https://github.com/user-attachments/assets/a5664859-aa78-4fe3-b5a1-17b7081fc353" />

love-linger:
> 高亮当前分支和我的这个分支线跟踪两码事

你的目的无非是将选中的路径与其他分别开来，这本身就是高亮分支的特性呀。

<img width="1173" height="1061" alt="Image" src="https://github.com/user-attachments/assets/7cb0c297-795a-4ecd-a1bc-41b039dfd1e9" />

[...]

PR #2646, opened 2026-08-22
[Enhancement] Add new graph-highlighting option "Selected Commits (only first-parent)"

goran-w:
This PR does some refactoring of the graph-highlighting code and adds a new highlighting-option variant "Selected Commits (only first-parent)".

The new highlighting-option is similar to the existing highlighting-option "Selected Commits Only", but restricted to only ever highlighting the first-parent when going backwards in history from the selected commit(s). This feature was first discussed in comments on a related issue [here](https://github.com/sourcegit-scm/sourcegit/issues/2336#issuecomment-4477907019).

I find it very helpful to restrict the highlighting to first-parent, while still drawing the other parents. I've implemented it in the easiest way I could find, which was to only apply it to "selected commits" in a new option variant. (It would be possible, but a bit more involved, to instead implement it as a separate "perpendicular" option that could be applied to the "All" and "Current branch" variants as well...)

NOTE: This new highlighting-option is somewhat related to the Git-option `--first-parent` (which is applied by SourceGit's show-flag "First-parent only"), but it's not the same! The new option only restricts the highlighting of non-first-parents, while the existing show-flag hides them entirely. (Also, the option and flag can be used simultaneously without issues.)

### DIFF
--- a/src/Models/CommitGraph.cs
+++ b/src/Models/CommitGraph.cs
@@ -13,6 +13,7 @@ namespace SourceGit.Models
         All = 0,
         CurrentBranchOnly,
         SelectedCommitsOnly,
+        SelectedCommitsOnlyFirstParent,
         CurrentBranchAndSelectedCommits,
     }
 
@@ -156,6 +157,7 @@ namespace SourceGit.Models
 
                     if (!isHighlighted &&
                         (highlighting == CommitGraphHighlighting.SelectedCommitsOnly ||
+                         highlighting == CommitGraphHighlighting.SelectedCommitsOnlyFirstParent ||
                          highlighting == CommitGraphHighlighting.CurrentBranchAndSelectedCommits))
                     {
                         isHighlighted = highlightExtraCommits.Remove(commit.SHA);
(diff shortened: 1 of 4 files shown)


JC-Chung:
`SelectedCommitsOnlyFirstParent` was inserted in the middle of the enum (between `SelectedCommitsOnly` and `CurrentBranchAndSelectedCommits`), shifting `CurrentBranchAndSelectedCommits` from `3` to `4`:

```csharp
public enum CommitGraphHighlighting
{
    All = 0,
    CurrentBranchOnly,
    SelectedCommitsOnly,
    SelectedCommitsOnlyFirstParent,   // new
    CurrentBranchAndSelectedCommits,  // 3 -> 4
}
```

Since `GraphHighlighting` is persisted as a raw int in `sourcegit.uistates`, users who had `CurrentBranchAndSelectedCommits` saved will silently get `SelectedCommitsOnlyFirstParent` after upgrading. Maybe append the new value at the end instead?

love-linger:
@JC-Chung good catch. I've pushed a new commit that fixes it.

goran-w:
> @JC-Chung good catch. I've pushed a new commit that fixes it.

For reference, this was done in commit e308279e9669c71be5187a9ae68cc5c8370b82fb. Excellent catch, indeed! Thanks!

PR #2648, opened 2026-08-23
Use standardized data directories on linux

mpeter50:
This pull request moves the data directory for linux users to the filesystem location where they will expect it.
General data in `~/.local/share/`, cached git username avatars in `~/.cache/`. Only the former is migrated.

I considered moving the config file to `~/.config/` but I find the "Open Data Storage Directory" function in the menu useful, and I was torn on whether I should point that to the config directory. Currently the data directory does not hold much, just the avatar caches, the config file, and a lock file (which should probably rather go to `/run/user/...`, but that depends on its exact purpose), but perhaps more things would be held in this directory, so I didnt want to risk making them less accessible.

Detailed reasons for implementing this changes can be read here: https://github.com/sourcegit-scm/sourcegit/pull/2088#issuecomment-5089628776

This PR is to supersede #2088 . Linux related functionality is handled in the separate Linux.cs file, with respect to the AppImage portable data function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,11 +72,11 @@ You can download the latest stable from [Releases](https://github.com/sourcegit-
 
 This software creates a folder, which is platform-dependent, to store user settings, downloaded avatars and crash logs.
 
-| OS      | PATH                                      |
-|---------|-------------------------------------------|
-| Windows | `%APPDATA%\SourceGit`                     |
-| Linux   | `~/.sourcegit`                            |
-| macOS   | `~/Library/Application Support/SourceGit` |
+| OS      | PATH                                             |
+|---------|--------------------------------------------------|
+| Windows | `%APPDATA%\SourceGit`                            |
+| Linux   | `~/.local/share/SourceGit`, `~/.cache/SourceGit` |
+| macOS   | `~/Library/Application Support/SourceGit`        |
 
 > [!TIP]
 > * You can open this data storage directory from the main menu `Open Data Storage Directory`.
--- a/src/App.axaml.cs
+++ b/src/App.axaml.cs
@@ -24,6 +24,8 @@ namespace SourceGit
         public static void Main(string[] args)
         {
             Native.OS.SetupDataDir();
+            Native.OS.SetupConfigDir();
+            Native.OS.SetupCacheDir();
 
             AppDomain.CurrentDomain.UnhandledException += (_, e) =>
             {
--- a/src/App.axaml.cs
+++ b/src/App.axaml.cs
@@ -24,7 +24,6 @@ namespace SourceGit
         public static void Main(string[] args)
         {
             Native.OS.SetupDataDir();
-            Native.OS.SetupConfigDir();
             Native.OS.SetupCacheDir();
 
             AppDomain.CurrentDomain.UnhandledException += (_, e) =>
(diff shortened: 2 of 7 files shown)


mpeter50:
If you agree to adding a new dependency, I would add `Xdg.Directories` like #2088 to simplify handling of the standard paths. Basically `GetXdgDir()` and its clunky string based handling could go away. I wanted to use enums, but as I found C# does not spport custom properties on enum variants, only in a very clunky way with reflection.

love-linger:

There's only one thing left: I keep to store the `process.lock` (used for singleton mode) in `${XDG_CACHE_HOME}/SourceGit/process.lock` instead of `${XDG_RUNTIME_HOME}/SourceGit/process.lock`. This is because some Linux distro has strict limitations for file in `${XDG_RUNTIME_HOME}`:

- File modes
- May be auto cleaned by the OS
- Must be updated per 6 hours
- ...

mpeter50:
Are you sure XDG_RUNTIME_HOME is automatically cleaned by some distro?

In fact, I would expect XDG_CACHE_HOME to be cleaned up periodically. The former is just a tmpfs, everything there is automatically lost on system shutdown. Can I ask which distro have you seen to be running cleanups on XDG_RUNTIME_HOME?

Edit: oh, indeed

> Files in this directory MAY be subjected to periodic clean-up. To ensure that your files are not removed, they should have their access time timestamp modified at least once every 6 hours of monotonic time or the 'sticky' bit should be set on the file.
> -- https://specifications.freedesktop.org/basedir/latest/#variables

mpeter50:
I believe the speciifc file mode of 700 only applies to the runtime directory itself:

> The directory MUST be owned by the user, and they MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700. 

On my system I have files with all kinds of permission modes set in there, even from modern, generally well behaved software like pipewire, dbus and systemd.

The auto cleanup needs to be taken care of, thats right. It kind of makes sense to auto delete unused files, to avoid getting the memory filled up over time, but I think the cleanup routines could at least check additionally whether the file is opened in any process. Because if it is, deletion wont reclaim any space anyway.

SourceGit would be using the XDG_RUNTIME_HOME directory very lightly, just a single file and that too always on the same path, so I think it would be okay if you just set the sticky bit on it. But it would still be good to delete the lock file on exit and crashes.

I think this is how its done in dotnet: https://learn.microsoft.com/en-us/dotnet/api/system.io.file.setunixfilemode?view=net-10.0

```c#
using System;
using System.IO;

string filePath = "/path/to/file";
File.SetUnixFileMode(filePath, File.GetUnixFileMode(filePath) | UnixFileMode.StickyByte);
```

love-linger:
Done. 

- On Windows
  - Portable mode: it stores in `data/process.lock`
  - Non-portable mode: it stores in `%APPDATA%/SourceGit/process.lock`
- On macOS, it stores in `~/Library/Application Support/SourceGit/process.lock`
- On Linux
  - Portable mode: it stores in `data/process.lock`
  - Non-portable mode: it stores in `$XDG_RUNTIME_DIR/sourcegit.instance.lock`

You can download the latest CI build from GitHub Action.

mpeter50:
@love-linger The release seems to work well, thank you! But I think it would be better if there were automatic config migration from the older `~/.local/share/SourceGit/` directory too, when `~/.sourcegit/` does not exist. I have done that manually for myself, but other people might appreciate it. Updates are manual on linux for now, so there are likely still people who still have their configs there. 

Also, I think #2088 can be closed now too.

Issue #962, opened 2025-02-10
[Enhancement] Searching file view does not filter view directly


Doodeletion:
Searching the file view does not filter the view immediately but rather displays a dropdown of matching files. You can only see the file contents when selecting a file from the dropdown. Once a file is selected, the file view will only show that one file. If you're searching for multiple similarly named files, you have to start a separate search for each one.

![Image](https://github.com/user-attachments/assets/10461d2b-07cc-4215-9e9f-c65861ae881c)

After selecting `RepositorySettings.cs` from the dropdown:

![Image](https://github.com/user-attachments/assets/fb2885d9-c4b2-4f0b-a4f9-0c9327c19969)

I think it would be better if the search filtered the view directly instead of searching a dropdown.

love-linger:
See #775

love-linger:
Imagine the following situation: In the case where you have just entered the first few characters, all the files may match the filter, so the number of files displayed is still too large. And currently, for the UI thread to not get stuck, SourceGit only shows the first 100 matching files.

mithom:
I would love for this issue to progress, at least in the discussion side.
would you be willing to look at suggested code implementations that take your concerns into account?
I'm not git wizard, but I know my way around performance. together maybe we can find create something working.

mithom:
i think the performance issue can be solved using 2 principles
* keep the scroll-box with a maximum of 100 files.
* when typing, instead of showing an empty tree list, show a message: 'make your search more specific to show the file tree
* when there are only 100 files or less matching the search. Create the tree, possibly collapsed to save resources when > 20 files

possible extra to discuss: make the 100 and/or 20 a setting the user can configure.

alternative approach:
* keep everything as it is now
* allow the user to select multiple files from the scroll-box by using shift/alt
* create the tree for all selected files instead of for only 1

Issue #775, opened 2024-12-02
Add Files search option


igor84:
As a long time Git Extensions user I love SourceGit. Might be the first real cross platform alternative I have found. Based on some initial testing I only detected two features I am missing for now:

1. I would like using Ctrl+f in Files tab of a selected commit to allow me to find a file by name. As far as I can see Git Extensions is using this command to list all the files: `git ls-tree -z -r --name-only <commit_hash>`.
2. In the changes tab I can search within a diff of selected file but I would like the search to automatically go through all the files in the diff.

love-linger:
duplicate #730

igor84:
I approve of your concerns in that ticket. I think Git Extensions goes around those problems by:

1. Not loading all the files unless user activates the file search using the Ctrl+F option
2. It just stores a list of all the files that it gets from the above `git ls-tree` command in memory in `List<string>`
3. In the UI it only shows a small popup with an input field and only after you type something into it it shows the first 20 matching results from that list as plain text paths (image attached)
4. As soon as that popup is dismissed it probably leaves that `List<string>` to the GC.

In Git Extension this is working flawlessly and with no visible increase in memory usage.

![image](https://github.com/user-attachments/assets/20e87a44-637a-44d5-a6a0-979696aae020)

love-linger:
You can download the latest CI build from Github Action

igor84:
Wau. Thanks for the quick reponse. I gave it a try on my ~90K files repo. When I open the repo SourceGit is taking ~90MB RAM, when I start typing in the Search Files box it jumps to ~180MB and after I finish it drops to ~150MB. The app is very responsive and does the search flawlessly so I think all this is ok.

As for UX, when I select a file from the drop down list the File List is filtered to only show the selected file while I expected it will still show the full tree with only the relevant folder expanded and the found file selected. I don't currently see anything really wrong with your solution, it is just different from what I am use to. Maybe it would be relevant in rare cases where for example I know all files related to some feature are in the same folder and I need some file A whose name I can't remember, but I do remember the name of file B that I know is in the same folder. I could then search for file B so that File List selects that file and then I could just see what other files are in the same folder.

love-linger:
I would like to use the same behaviour with other search implementation, such as `Search Changes` and `Search Branches/Tags/Submodules`, for this feature.

Issue #730, opened 2024-11-21
[Commit Files Tab] Add a files search bar


albertodlc:
![image](https://github.com/user-attachments/assets/3356d9e1-78d6-492a-ac7a-550dcdda45d4)

love-linger:
The revision file system tree in the `FILE` page is not calculated once. 

Instead, the first layer is calculated when the current preview commit/revision changes. And when the user expands a folder node for the first time, the next layer is calculated.

The advantages of this approach are:

* It is very fast to loading data when user changes the preview commit, basically avoids the UI lag
* Uses less memory since we just load the data we wanted
* Avoid system gc for very large repositories

For example: the `UnrealEngine` repository has 183K+ files. If we load all datas, and build the entire filesystem tree for it:

* It will take very long time to build this tree.
* It will cost very large memory to hold the datas.
* The system gc may be triggered very frequently.

So, we can not search on this tree because we do not have enough data to do this job.

love-linger:
One more thing, I think search commit by file path is another way to do this. 
![image](https://github.com/user-attachments/assets/353bc432-5d63-4aa9-98e5-8779991e013c)

PR #2654, opened 2026-08-26
ux: preview the selected file when navigating file search results

weiningwei:
## Summary

Improves the file search experience in the commit details **Files** tab. Previously, picking a result from the search dropdown only wrote the full path back into the search box and filtered the file tree down to a single file — but the tree node was not selected, so the file content was not shown until the user clicked the file again in the tree.

With this change:

- Navigating the search dropdown with the arrow keys **previews the selected file directly** in the file tree and the content viewer (the tree node is now selected automatically).
- The search keyword is **no longer overwritten** when picking a suggestion, so multiple similarly-named files can be reviewed without re-searching.

## Related issues

I'm aware of the earlier discussions in #730, #775, and especially #962 ("Searching file view does not filter view directly"). The main concern raised there was performance / lazy-loading when filtering the whole tree. This PR deliberately keeps the existing dropdown approach and only ever renders a **single file** in the tree, so it does not introduce the large-repository / UI-freeze concern discussed in those issues.

## Changes

- `Views/RevisionFileTreeView.axaml(.cs)`: select the matched node after rebuilding the tree so the content viewer loads it.
- `Views/RevisionFiles.axaml(.cs)`: hook the dropdown `SelectionChanged` to live-preview the selected item; stop overwriting the search filter; close the dropdown and return focus to the search box on click/Enter.
- `ViewModels/CommitDetail.cs`: drop the exact-path exclusion in the matcher (it was only needed for the old overwrite-the-filter behavior).

5 files changed, 37 insertions(+), 11 deletions(-).

### DIFF
--- a/src/ViewModels/CommitDetail.cs
+++ b/src/ViewModels/CommitDetail.cs
@@ -662,8 +662,7 @@ namespace SourceGit.ViewModels
             var suggestion = new List<string>();
             foreach (var file in _revisionFiles)
             {
-                if (file.Contains(_revisionFileSearchFilter, StringComparison.OrdinalIgnoreCase) &&
-                    file.Length != _revisionFileSearchFilter.Length)
+                if (file.Contains(_revisionFileSearchFilter, StringComparison.OrdinalIgnoreCase))
                     suggestion.Add(file);
 
                 if (suggestion.Count >= 100)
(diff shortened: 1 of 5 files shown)


love-linger:
I do not want to add the preview feature when selection changed in suggestion dropdown. I've pushed a commit that only auto-select the file node in tree after picking someone in the suggestion dropdown.

weiningwei:
Thanks for the feedback. I'll close this PR.

Issue #2656, opened 2026-08-26
[UX] Add Ctrl+Ins copy shortcut in diff view for Windows


bartizan:
On Windows/Linux, the diff view supports `Ctrl + C` for copy, but unlike other text inputs and the standard Windows convention it does not support `Ctrl + Ins`.

love-linger:
Merged

PR #2657, opened 2026-08-26
ux: add Ctrl+Ins copy shortcut in diff view

bartizan:
https://github.com/sourcegit-scm/sourcegit/issues/2656

### DIFF
--- a/src/Views/TextDiffView.axaml.cs
+++ b/src/Views/TextDiffView.axaml.cs
@@ -662,7 +662,7 @@ namespace SourceGit.Views
         {
             if (e.KeyModifiers.Equals(OperatingSystem.IsMacOS() ? KeyModifiers.Meta : KeyModifiers.Control))
             {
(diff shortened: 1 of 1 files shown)


src/Views/TextDiffView.axaml.cs
```diff
@@ -662,7 +662,7 @@ private async void OnTextAreaKeyDown(object sender, KeyEventArgs e)
         {
             if (e.KeyModifiers.Equals(OperatingSystem.IsMacOS() ? KeyModifiers.Meta : KeyModifiers.Control))
             {
-                if (e.Key == Key.C)
+                if (OperatingSystem.IsMacOS() ? e.Key == Key.C : e.Key is Key.C or Key.Insert)
```
love-linger: I suggest to just remove the `OperatingSystem.IsMacOS`.

bartizan: fixed

PR #2666, opened 2026-08-30
  ux: add tooltip to copy button in Launcher notification

weiningwei:
## What

Shows a transient checkmark icon as visual feedback when the user copies a notification message, then resets back to the copy icon after one second.

## Changes

- Make `Notification` observable (`ObservableObject`) and add an `IsCopied` property.
- Toggle between the copy icon and a green checkmark in the launcher notification item based on `IsCopied`.
- Add a `DispatcherTimer` in `LauncherPage` to reset `IsCopied` one second after copying.
- Add a tooltip to the copy button.

### DIFF
--- a/src/Views/LauncherPage.axaml
+++ b/src/Views/LauncherPage.axaml
@@ -166,7 +166,7 @@
                         <Path Grid.Column="0" Width="14" Height="14" Data="{StaticResource Icons.Info}" Fill="Green" IsVisible="{Binding !IsError}"/>
                         <TextBlock Grid.Column="1" Margin="8,0,0,0" FontWeight="Bold" FontSize="14" Text="{DynamicResource Text.Launcher.Error}" IsVisible="{Binding IsError}"/>
                         <TextBlock Grid.Column="1" Margin="8,0,0,0" FontWeight="Bold" FontSize="14" Text="{DynamicResource Text.Launcher.Info}" IsVisible="{Binding !IsError}"/>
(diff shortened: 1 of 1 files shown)


JC-Chung:
Hi @weiningwei, thanks for this PR! I ran into the same need for `CommitBaseInfo`'s SHA copy button and, as a follow-up, ended up building a reusable `CopyButton` control (#2668), covering both `LauncherPage` and `CommitBaseInfo`. It touches the same lines in `LauncherPage.axaml(.cs)`, so merging both would conflict. Would you be open to rebasing this on top of `CopyButton` so we land on one consistent solution? Happy to help adjust either side.

weiningwei:
Hi @JC-Chung, congrats on #2668 being merged — the reusable `CopyButton` is a nice cleanup, and thanks for the heads-up here.
I've rebased this PR on top of it. Since the notification copy feedback is now fully covered by `CopyButton`, this PR is reduced to a single line: adding the missing `Text.Copy` tooltip to the `CopyButton` in `LauncherPage` — the same tooltip `CommitBaseInfo` already has, so both usages stay consistent.
@love-linger Just want to confirm: is this one-liner worth keeping as a separate PR, or would you rather not have the tooltip there at all? Happy to close if it's unnecessary.

JC-Chung:
I did notice the tooltip aspect, but since the dismiss button next to it doesn't have one either, I kept it as-is. If we're adding one now, maybe we should also consider adding a tooltip to the dismiss button? No rush to change it right away though — might be worth waiting to hear what the @love-linger think first.

PR #2668, opened 2026-09-01
enhance: extract reusable CopyButton control with copy feedback

JC-Chung:
## What
Extracts the "copy + show checkmark for 1s" logic into a reusable
`CopyButton` control (`src/Views/CopyButton.cs`), and uses it in:
- `CommitBaseInfo` (copy commit SHA)
- `LauncherPage` (copy notification message)

## Why
Both places previously implemented very similar copy-feedback logic
(a `DispatcherTimer` + two icons toggled by an `IsCopied` flag)
separately. `CopyButton` consolidates this into one control driven by
a `CopyText` property, so no changes to data models are needed.

## Note
This overlaps with #2666, which adds the same copy-feedback UX to
`LauncherPage` via `Notification.IsCopied`. Left a comment there to
discuss which approach to land on.

### DIFF
--- a/src/Views/CopyButton.cs
+++ b/src/Views/CopyButton.cs
@@ -13,13 +13,16 @@ namespace SourceGit.Views
     {
         protected override Type StyleKeyOverride => typeof(Button);
 
-        public static readonly StyledProperty<string> CopyTextProperty =
-            AvaloniaProperty.Register<CopyButton, string>(nameof(CopyText), string.Empty);
+        public static readonly DirectProperty<CopyButton, string> CopyTextProperty =
+            AvaloniaProperty.RegisterDirect<CopyButton, string>(
+                nameof(CopyText),
+                static o => o.CopyText,
+                static (o, v) => o.CopyText = v);
 
         public string CopyText
         {
-            get => GetValue(CopyTextProperty);
-            set => SetValue(CopyTextProperty, value);
+            get => _copyText;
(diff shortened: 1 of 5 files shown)


src/Views/CopyButton.cs
```diff
@@ -0,0 +1,124 @@
...
+using Avalonia.Media;
+using Avalonia.Threading;
+
+namespace SourceGit.Views
+{
+    public class CopyButton : Button
+    {
+        protected override Type StyleKeyOverride => typeof(Button);
+
+        public static readonly StyledProperty<string> CopyTextProperty =
```
love-linger: `CopyText` is a per-instance-data. I suggest to use `DirectProperty` instead of `StyledProperty` for it.

JC-Chung: Pushed two commits addressing this:
1. `fix: ...` — `OnClick` was setting `IsCopied = true` even when `CopyText` was empty, so it now returns early in that case. Also removed the now-redundant `if (_resetTimer != null)` check in `OnUnloaded`, since `_resetTimer` is always initialized in `OnLoaded` by the time `OnUnloaded` runs (same pattern as `CommitBaseInfo`'s `_iconResetTimer`, which never null-checks it either).
2. `code_style: ...` — switched `CopyText` to `DirectProperty` as suggested, since it's purely per-instance data with no styling/animation use case, matching the existing `IsCopied` property.

PR #2669, opened 2026-09-01
show parent folder name on tabs for repositories with duplicate names

flowprog:
## Description

When multiple repositories with the same name are opened at the same time, their tabs are indistinguishable — you can only tell them apart by switching through them one by one.

This PR makes the tab bar disambiguate duplicate names by appending the parent folder name:

- unique name: `repo`
- duplicated name: `repo (parent-folder)`

## Implementation

- Add a `DisplayName` property to `LauncherPage` (initialized to the node name) and bind the tab text to it instead of `Node.Name`.
- Add `Launcher.RefreshTabDisplayNames()`, which counts repository tab names and appends the parent folder (from the repository path) only when a name is used by more than one open tab.
- Refresh display names whenever the tab set can change: opening a repository/sub-repository, closing a tab / other tabs / tabs on the right, and renaming a repository node in preferences.

No behavior change for uniquely-named tabs.

### DIFF
--- a/src/ViewModels/EditRepositoryNode.cs
+++ b/src/ViewModels/EditRepositoryNode.cs
@@ -48,6 +48,7 @@ namespace SourceGit.ViewModels
             {
                 Preferences.Instance.SortByRenamedNode(_node);
                 Welcome.Instance.Refresh();
+                App.GetLauncher().RefreshTabDisplayNames();
             }
 
             return Task.FromResult(true);
--- a/src/ViewModels/Launcher.cs
+++ b/src/ViewModels/Launcher.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Text;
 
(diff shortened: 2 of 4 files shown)


love-linger:
I do not think this is the right way to distinguish two repository with the same name.

- `SourceGit` allows user to create custom groups to manage repositories. So is it better to use group name instead of the parent folder name in disk?
- `SourceGit` allows user to give a friendly name for repository (not the folder name, just a label in tab. For example, you can rename two repositories with `REPO (fork)`, `REPO (github)`
- `SourceGit` allows user to set different bookmark for repositories with the same name.

flowprog:
You're referring to a workspace when you say "group," right? Often, duplicate names happen because they come from the same project—a server-side project, a client-side project, or a front-end project—and they're usually placed in the same workspace.

Being able to fully customize names is great, but users need to manually modify them. If the goal is to minimize extra modifications and increase the burden while still providing intuitive differentiation, then automatic naming is also an option. Frankly, I wouldn't have noticed the option to modify it until you mentioned it.

Of course, this is just my opinion.

love-linger:
No, this is group:

image

PR #2678, opened 2026-09-04
feat: responsive layouts and optional split views

TrainerHol:
- Add responsive layouts for narrow and tiled windows.
- Allow History, Local Changes, and Stashes to be displayed in a persistent secondary pane.
- Support side-by-side and stacked split orientations.
- Allow repository view navigation to be placed in the sidebar or at the top.

Current View Tiled:
image

After:

Stacked View Tiled:
image
Side-by-side View Tiled:
image

### DIFF
--- a/src/Models/RepositoryUIStates.cs
+++ b/src/Models/RepositoryUIStates.cs
@@ -7,6 +7,12 @@ using Avalonia.Collections;
 
 namespace SourceGit.Models
 {
+    public enum RepositoryWorkspaceOrientation
+    {
+        SideBySide,
+        Stacked,
+    }
+
     public class RepositoryUIStates
     {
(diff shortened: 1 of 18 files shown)


love-linger:
- Too many files have been modified, and the changes do not serve the same purpose. I cannot conduct the review at all (I have no idea what each change is intended for).
- Judging from the problematic screenshot above, the MinWith constraint is the cause. Your relevant modifications were meant to ensure proper content display when the window is too small. However, you also added the Split Views feature. While it seems to improve space‑utilization, each view becomes too tiny. For frequently‑used functions such as diff, it is practically unusable. You may also search the project’s ISSUES; numerous reports complain about the excessively small diff view.

PR #2679, opened 2026-09-05
ux: focus input-less popups

corebonts:
…so Space/Enter in Discard confirmation dialog no longer stages the file behind it.

Previously the focus was left behind in the main window, which could lead to accidental staging or commit.


love-linger:
I've pushed my fix for this BUG.

corebonts:
https://github.com/sourcegit-scm/sourcegit/commit/e9d7d56bf035704b409598da57f221dab699e454

PR #2687, opened 2026-09-09
feat: support custom HTTP headers in AI service configuration

### LINKED ISSUE #2686: Support custom HTTP headers in AI service configuration (required for OpenCode Go API)
## Description

When configuring SourceGit's AI chat to use the [OpenCode Go API](https://opencode.ai/docs/go/) endpoint, the request fails with HTTP 400 and a `MissingSessionID` error.

OpenCode Go now requires an `x-opencode-session` header on every request for routing optimization. Without this header, the API returns:

```
HTTP 400 (MissingSessionID: )

Error from provider (Console Go): Request is missing x-opencode-session
and cannot be routed efficiently.
Please see https://opencode.ai/docs/go/#where-can-i-use-it
```

## Expected Behavior

SourceGit should either:

1. Allow users to configure custom HTTP headers for AI services (e.g., via the settings UI or configuration file), **or**
2. Automatically detect and include the `x-opencode-session` header when using OpenCode Go endpoints (e.g., when the server URL matches `opencode.ai`).

## Current Behavior

There is no way to add custom HTTP headers to AI service requests. The `AI.Service` model (`src/AI/Service.cs`) only supports `Server`, `ApiKey`, and `Model` configuration. The OpenAI .NET SDK already supports custom headers via `OpenAIClientOptions.DefaultHeaders`, but SourceGit does not expose this capability.

## How to Reproduce

1. Go to **Preferences > AI Services** and add a new service.
2. Set the **Server URL** to `https://opencode.ai/zen/go/v1`
3. Enter a valid API key and select a model.
4. Attempt to use AI chat or generate a commit message.
5. Observe the HTTP 400 `MissingSessionID` error.

## Proposed Solution

Add a **Custom Headers** field (key-value pairs) to the AI service configuration in `src/AI/Service.cs`. This would allow users to set headers like:

| Key                    | Value          |
|------------------------|----------------|
| `x-opencode-session`   | `<session-id>` |

In the code, the headers can be applied when constructing the `OpenAIClient`:

```csharp
var options = new OpenAIClientOptions()
{
    Endpoint = new Uri(Server),
};

// Apply custom headers
foreach (var header in CustomHeaders)
    options.DefaultHeaders.Add(header.Key, header.Value);

return new OpenAIClient(credential, options);
```

This approach is flexible and would benefit any OpenAI-compatible API provider that requires custom headers, not just OpenCode Go.

## Environment

- SourceGit version: 2026.19 (latest)
- OS: macOS / Windows / Linux
- AI Provider: OpenCode Go API

yopichy:
## Summary

Adds support for custom HTTP headers in the AI service configuration. This fixes #2686.

Some OpenAI-compatible providers require extra headers that SourceGit previously had no way to send. For example, OpenCode Go now requires an `x-opencode-session` header on every request and rejects calls without it:

```
HTTP 400 (MissingSessionID: )
Error from provider (Console Go): Request is missing x-opencode-session
and cannot be routed efficiently.
```

## Changes

- `src/AI/Service.cs`
  - New `ExtraHeaders` property (`Dictionary<string, string>`, persisted in preferences JSON; old configs without it default to empty).
  - New `ExtraHeadersText` facade for UI editing: one `Name: Value` pair per line; lines without a colon or with an empty/invalid name are ignored; names are case-insensitive.
  - Headers are injected via a small `System.ClientModel` `PipelinePolicy` (`PerCall`) in `GetOpenAIClient()`, so they apply to both `OpenAIClient` and `AzureOpenAIClient` — covering AI chat, commit-message generation, and model-list fetching.
- `src/Views/Preferences.axaml` — new "Extra Headers" multi-line field in the AI service editor.
- `src/Resources/Locales/en_US.axaml`, `id_ID.axaml` — label string (other locales fall back to the existing localization-bot sync).

## How to use (OpenCode Go example)

In **Preferences > AI Services**, add to the **Extra Headers** field:

```
x-opencode-session: <a-stable-uuid>
```

## Evidence

AI Assistant successfully generating a commit message through the OpenCode Go endpoint with the custom header set (model `mimo-v2.5`, token usage reported normally instead of the previous HTTP 400):

AI Assistant generating a commit message via OpenCode Go with x-opencode-session header

## Verification

- `dotnet build src/SourceGit.csproj -c Release` → succeeded, 0 warnings, 0 errors.
- Runtime check of the header parsing: valid pairs parsed, malformed lines ignored, case-insensitive de-duplication (last wins), text and JSON round-trips verified, empty default confirmed.

### DIFF
--- a/src/AI/Service.cs
+++ b/src/AI/Service.cs
@@ -1,7 +1,10 @@
 ﻿using System;
 using System.ClientModel;
+using System.ClientModel.Primitives;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text.Json.Serialization;
+using System.Threading.Tasks;
 using Azure.AI.OpenAI;
 using CommunityToolkit.Mvvm.ComponentModel;
 using OpenAI;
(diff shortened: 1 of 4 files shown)


love-linger:
Merged with some changes:

- Pick this commit to `develop` branch instead of `master`
- Code cleanup
- Add Chinese (Simplified) translation

PR #2689, opened 2026-09-10
feat(ui): show dirty state indicator in pages command palette

JC-Chung:
## What
Add a DirtyState indicator dot and tooltip to each open tab (LauncherPage) in the Pages Command Palette (Ctrl+P / ⌘+P), matching the presentation already used in LauncherTabBar.

## Changes
- `LauncherPagesCommandPalette.axaml`:
  - Add `xmlns:m="using:SourceGit.Models"` namespace
  - Add an `Ellipse` in the `vm:LauncherPage` DataTemplate that reflects `DirtyState` via `DirtyStateConverters.ToBrush`
  - Add `ToolTip.Tip` showing the repository path (`Node.Id`) and `DirtyState` description (`DirtyStateConverters.ToDesc`); the welcome tab still shows its original text

## Why
Lets users see at a glance whether a tab has uncommitted changes or is out of sync with upstream, without having to switch to it first.

### DIFF
--- a/src/Views/LauncherPagesCommandPalette.axaml
+++ b/src/Views/LauncherPagesCommandPalette.axaml
@@ -2,6 +2,7 @@
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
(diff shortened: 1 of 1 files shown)


love-linger:

image

JC-Chung:
> I've pushed my implementation for this feature.

This ahead/behind formatting logic already exists in `Branch` and `RepositoryStatus`, and now a third copy here — worth extracting into a common helper?

Issue #2716, opened 2026-09-15
Weird window position on Linux


maxsnts:
Hello

In Linux (Fedora gnome), when docking the window to half the screen (or any position near the screen edge), the window does not take the available space.
Any ideia why?

<img width="1308" height="169" alt="Image" src="https://github.com/user-attachments/assets/227cf339-08f7-4eda-b184-924836bc7c67" />

Not a "problem", just curious. Is this intentional?

Thank you

Best Regards
Max

gadfly3173:
The part that looks blank is the window shadow. Because of Avalonia, this shadow takes up space within the window itself.

PR #2717, opened 2026-09-16
fix: collapse shadow padding when window touches a screen edge

gadfly3173:
Closes #2716


dwndland:
Tested on Endeavour OS)

It might be a little special, but it does not work in all the cases.
In my setup, two monitors are alligned vertically, so Monitor 2 is above Monitor 3.
And all monitors have a bottom panel configureed.
The shadow disappears at the bottom panel only if there is no screen below, so on my upper screen it still snaps with this shadow gap.

(Just as a side note. My setup is special, so this would not block anything IMHO)

gadfly3173:
Avalonia can detect each physical monitor through `Screen.Bounds`, but `Screen.WorkingArea` is the usable area after panels and is not always available per monitor on Linux/X11.  
With vertically stacked monitors, the upper monitor’s bottom panel is an internal edge of the virtual desktop, so the window may still retain its bottom shadow there:

```text
┌──────────────────┐
│ Upper monitor    │
│                  │
│                  │
│ Upper panel      │
└──────────────────┘  ← internal desktop edge
┌──────────────────┐
│ Lower monitor    │
│                  │
│                  │
│ Lower panel      │
└──────────────────┘
```

This comes from `src/Avalonia.X11/Screens/X11Screen.Providers.cs`, where `UpdateWorkArea()` reads the root `_NET_WORKAREA` and sets `WorkingArea = Bounds.Intersect(wa)`; `src/Avalonia.Controls/Platform/ScreenHelper.cs` then selects a screen based on the window’s intersecting area.

dwndland:
I am on wayland, just as a side note :D
I think its right what you say. The working area does not support that something in the vertical middle of it is stealing space. I guess.

But interestingly, this is he only exception its seems.
My setup is following.
I have 3 monitors horizontally and a 4th in the center on top.
The screen edges left and right of the top monitor is seen right, just the lower panel not. I would have expected the working area would be have an imaginary monitor top left and top right, to get a full square :D

